PR #110, opened 2019-10-18
Moving from assert --> throw.

s-mandra:
This PR should remove all the asserts from the code. It suffices to change:
```
assert(msg) --> throw ERROR_MSG(msg)
```
I notices that tests use `EXPECT_DEATH` which should changed to `EXPECT_ANY_THROW`.


s-mandra:
See #204.

PR #180, opened 2019-10-24
Fix Autoconf.

s-mandra:


README.md
```diff
@@ -34,8 +34,8 @@ $ autoreconf -i && autoconf && ./configure
 $ make && make run-tests && make install
 ```
 
-To disable tests based on `cirq`, use `./configure --disable-cirq_tests`
-instead. By default, qFlex is installed in `$HOME/local/`. To change the
+To disable qFlex python interface, use `./configure --disable-pybind11`.
```
95-martin-orion: Is the flag `--disable-pybind` or `--disable-pybind11`? Both appear in this PR.

configure.ac
```diff
@@ -104,85 +103,75 @@ AC_ARG_ENABLE([openmp],
...
-AS_IF([test ${openmp} = "true"], AC_MSG_WARN([OpenMP is still experimental with potential memory leak.]))
-
-AC_SUBST(openmp, ${openmp})
+AS_IF([test ${openmp} == "true"], 
+  AC_CHECK_HEADER([omp.h], [], [AC_MSG_ERROR([Please install omp.h before configuring.])])
+  AC_MSG_WARN([OpenMP is still experimental with potential memory leak.])
+)
 
-# Check for Python version
+# Check for python. If --disable-pybind, no reasons to activate python
```
95-martin-orion: `--disable-pybind` here (vs. `--disable-pybind11` in Makefile)

configure.ac
```diff
@@ -104,85 +103,75 @@ AC_ARG_ENABLE([openmp],
...
+CIRQ_REQ_VERSION="0.5"
+CIRQ_ERR_FLAG=pybind11
+
+PYBIND11_REQ_VERSION="2.2.4"
+PYBIND11_ERR_FLAG=pybind11
+
+PYTEST_REQ_VERSION="4.5"
+PYTEST_ERR_FLAG=python_tests
+
+# Check python modules. If --disable-pybind, no reasons to activate python
```
95-martin-orion: `--disable-pybind` here as well.

configure.ac
```diff
@@ -104,85 +103,75 @@ AC_ARG_ENABLE([openmp],
...
+if test ${pybind11} == "true" && ${python_checks} == "true"; then
+
+  python_check_module cirq     ${CIRQ_REQ_VERSION}     ${CIRQ_ERR_FLAG}
+  python_check_module pybind11 ${PYBIND11_REQ_VERSION} ${PYBIND11_ERR_FLAG}
+  python_check_module pytest   ${PYTEST_REQ_VERSION}   ${PYTEST_ERR_FLAG}
+
+fi
+
+# If --disable-python, switch off everything
```
95-martin-orion: Is `--disable-python` distinct from `--disable-pybind / --disable-pybind11`?

docs/install.md
```diff
@@ -38,9 +38,10 @@ $ autoreconf -i && autoconf && ./configure --help
 
 The following options for `configure` are also available:
 ```
---enable-openmp       Enable experimental support for OpenMP
---disable-pybind11    Disable installation of Python porting of qFlex
---disable-cirq_tests  Disable tests which depends on cirq
+--enable-openmp           Enable experimental support for OpenMP
+--disable-python_checks   Disable python checks while configuring
```
95-martin-orion: Is `python_checks` necessary as a separate flag? It seems that we should always run these checks if pybind is enabled.

docs/install.md
```diff
@@ -38,9 +38,10 @@ $ autoreconf -i && autoconf && ./configure --help
...
+--disable-pybind11        Disable installation of Python porting of qFlex
+--disable-python_tests    Disable python tests
```
95-martin-orion: RE `--disable-python_tests`: I'd recommend using either all dashes or all underscores in flag names, to make them easier to remember.

README.md
s-mandra: It's `--disable-pybind11`. Fixed everywhere.

configure.ac
s-mandra: Fixed.

configure.ac
s-mandra: Fixed :)

configure.ac
s-mandra: Fixed. There should be only the following flags:
```
--disable-python_checks        Disable only checks while configuring
--disable-python_tests            Disable only python tests
--disable-pybind11                   Disable pybind11 (with all the checks and tests)
```
Thanks for catching these!

docs/install.md
s-mandra: > Is `python_checks` necessary as a separate flag? It seems that we should always run these checks if pybind is enabled.

It's useful when you're installing containers. For instance, you can start building your container without all the Python dependencies and adding them only when required (as we do for the different layers in `docker`).

docs/install.md
s-mandra: > RE `--disable-python_tests`: I'd recommend using either all dashes or all underscores in flag names, to make them easier to remember.

I cannot :( This is a limitation of `autoconf`. `autoconf` identifies everything after `--disable-{variable_name}` as variable name. If I try `--disable-python-tests` gives me an error since I cannot create a variable called `python-tests`. That's the reason why I need underscores. I can only change to `python-tests --> pythontests`.

95-martin-orion: Unfortunate, but I guess those familiar with autoconf would also be aware of this restriction. Thanks for checking!

README.md
```diff
@@ -60,6 +60,10 @@ To disable qFlex python interface, use `./configure --disable-pybind11`.
 By default, qFlex is installed in `$HOME/local/`. To change the
 installation folder, use `./configure --prefix=/new/installation/folder/`. 
 
+After running these commands, qFlex can be installed by running `make install`.
```
95-martin-orion: Duplicated lines?

s-mandra: Good eye! :) Yet, it was duplicated. Thanks!

PR #189, opened 2019-10-25
Removing asserts

alankao64:
Continuing #110.


95-martin-orion:
Interesting - this appears to be a real, globally-visible branch, as opposed to the temp branches seen in tests. Any idea what changed?

alankao64:
I thought that in the tests we were using accounts that were not part of the qFlex developing team, where as this account is part of the team so allowed to push branches.

src/tensor.cpp
```diff
@@ -765,17 +764,17 @@ size_t Tensor::num_zeros() const {
   return count;
 }
 
-void Tensor::print() const {
-  std::string print_str("Tensor of rank ");
-  print_str += std::to_string(_indices.size());
-  print_str += ": ";
+std::string Tensor::tensor_to_string() const {
```
95-martin-orion: Since we also have `print_data`, calling this something like `summary_string` might help distinguish between the two.

alankao64:
ran into merge issues, new branch assert_removal will take care of this PR.

s-mandra:
See #204.

Issue #191, opened 2019-10-26
Improve Python Quality


s-mandra:
We are graded B :(

alexandrupaler:
i will take care of this

alexandrupaler:
Code quality is A now. Closing the issue.

Issue #177, opened 2019-10-24
Flake in Cirq interface


95-martin-orion:
Tempfile close is racing with object destructors, causing a flake.

alexandrupaler:
i will take care of this.

alexandrupaler:
Related by the folder restructuring of #249?

Issue #174, opened 2019-10-24
mkstemp issues in pytest


s-mandra:
```
python3 -m pytest tests/python/simulation_test.py 
Test session starts (platform: linux, Python 3.7.5, pytest 5.2.1, pytest-sugar 0.9.2)
rootdir: /home/dagart/NASA_Local/repository/qflex
plugins: sugar-0.9.2
collecting ... 
 tests/python/simulation_test.py ✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓                                         100% ██████████
===================================================================================== warnings summary ======================================================================================
/home/dagart/.local/lib/python3.7/site-packages/cirq/google/engine/engine.py:32
  /home/dagart/.local/lib/python3.7/site-packages/cirq/google/engine/engine.py:32: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Iterable

-- Docs: https://docs.pytest.org/en/latest/warnings.html

Results (6.37s):
     100 passed
Exception ignored in: <function QFlexCircuit.__del__ at 0x7fe311b9f200>
Traceback (most recent call last):
  File "/home/dagart/NASA_Local/repository/qflex/python/cirq_interface/qflex_circuit.py", line 84, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function QFlexOrder.__del__ at 0x7fe311b98f80>
Traceback (most recent call last):
  File "/home/dagart/NASA_Local/repository/qflex/python/cirq_interface/qflex_order.py", line 52, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <function QFlexGrid.__del__ at 0x7fe311d19e60>
Traceback (most recent call last):
  File "/home/dagart/NASA_Local/repository/qflex/python/cirq_interface/qflex_grid.py", line 60, in __del__
TypeError: 'NoneType' object is not callable
```

kkissell:
This is part of what I believe I fixed in the patch in #289

[...]

PR #195, opened 2019-10-28
improve python quality

alexandrupaler:
solves #191

### DIFF
--- a/python/cirq_interface/data_storage_interface.py
+++ b/python/cirq_interface/data_storage_interface.py
@@ -20,7 +20,7 @@ class DataStorageInterface:
 
         try:
             # https://stackoverflow.com/questions/36069102/pytest-fixture-finalizer-typeerror-nonetype-object-is-not-callable
-            yield
+            # yield
             os.close(self._file_handle[0])
         except OSError as e:
             if e.errno == 9:
--- a/python/cirq_interface/data_storage_interface.py
+++ b/python/cirq_interface/data_storage_interface.py
@@ -1,0 +1,35 @@
+import tempfile
+import os
+
+class DataStorageInterface:
+    def __init__(self, use_temp_files = True):
+
+        self.use_files = use_temp_files
+
+        if self.use_files:
+            # Behind the scene, this class creates a temporary file for each object
+            self._file_handle = tempfile.mkstemp()
+        else:
+            print("not implemented")
+
+    def __del__(self):
+
+        # The destructor removes the temporary file
+        if not self.use_files:
+            return
+
+        try:
+            # https://stackoverflow.com/questions/36069102/pytest-fixture-finalizer-typeerror-nonetype-object-is-not-callable
+            yield
+            os.close(self._file_handle[0])
+        except OSError as e:
+            if e.errno == 9:
+                # if it was closed before
+                pass
+            else:
+                raise e
+        except:
+            raise ValueError("Something is very wrong!")
+
+        # remove the temporary file from disk
+        os.remove(self._file_handle[1])
--- a/python/cirq_interface/data_storage_interface.py
+++ b/python/cirq_interface/data_storage_interface.py
@@ -1,8 +1,10 @@
 import tempfile
 import os
 
+
 class DataStorageInterface:
-    def __init__(self, use_temp_files = True):
+
+    def __init__(self, use_temp_files=True):
 
         self.use_files = use_temp_files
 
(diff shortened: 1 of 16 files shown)


python/cirq_interface/fsim_gate.py
```diff
@@ -25,9 +25,10 @@
 import math
 from typing import Optional
 
+import cirq
```
95-martin-orion: I think the warning on this line is because we also have `from cirq ...` further down. We should remove the `from cirq` imports and update their call sites to use the full `cirq.{packages}...` path.

alexandrupaler: I keep it like this and in the next PR move to 0.6 Cirq

alexandrupaler:
should solve #174 and #177 . This PR also prepares the way to remove temp files. Added data_storage_interface.py

python/cirq_interface/data_storage_interface.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __del__(self):
+
+        # The destructor removes the temporary file
+        if not self.use_files:
+            return
+
+        try:
+            # https://stackoverflow.com/questions/36069102/pytest-fixture-finalizer-typeerror-nonetype-object-is-not-callable
+            yield
```
95-martin-orion: Could you clarify how this works? My reading of the linked page (plus the ["docs" link](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code) on that page) is that this `yield` --> `os.close()` structure would need to be contained in a `pytest.fixture` method to behave as expected. (More info on how to use pytest.fixture is also given at that link)

python/cirq_interface/data_storage_interface.py
```diff
@@ -0,0 +1,35 @@
+import tempfile
+import os
+
+class DataStorageInterface:
```
95-martin-orion: Just a question, no change required: is the intent to later modify this class to support both temp-file and non-temp-file behavior?

tests/python/simulation_test.py
```diff
@@ -596,13 +594,13 @@ def test_simulation_with_fsim_gates(x):
         'final_state': final_conf
     }
 
-    print(options)
+    # print(options)
```
95-martin-orion: remove comment?

tests/python/qflex_circuit_test.py
```diff
@@ -13,6 +13,42 @@
...
+                             dummy_device_1,
+                             dummy_order,
+                             allow_decomposition=True)
+
+    qflexcirc2 = QFlexCircuit(circuit,
+                             dummy_device_1,
+                             dummy_order,
+                             allow_decomposition=True)
+
+    # should be ok
```
95-martin-orion: This comment can be removed, as it's implied by `assert`. Same below.

python/cirq_interface/data_storage_interface.py
alexandrupaler: Yes

python/cirq_interface/data_storage_interface.py
alexandrupaler: Short explanation: pure luck that it works. Elaborated (potential explanation): The yield transforms the method into a generator, and this fact solves a race condition in the destruction of the object.

alexandrupaler:
174 and 177 are solved only because of the yield in the destructor. Please keep the PR like this, if it is not too much to ask, because there is already another PR to remove temp files by Salvatore, and we would duplicate/overlap too much for simple things.

python/cirq_interface/data_storage_interface.py
95-martin-orion: As far as I can tell, the link in the comment is specific to `pytest.fixture`, and isn't related to this code - searching for "python yield destructor" turns up zero relevant results. Even if this appears to resolve the issue, we shouldn't rely on it.

alexandrupaler: The link in the comment mentions the yield as a solution that has applicability to the fixture.

alexandrupaler:
I am not in favour of removing files altogether, but I think I am not the one to decide. However, the version of the code three weeks ago had no temp files, then temp files appeared and introduced very strange behaviours which took days to debug and understand, and I had to write my code around them. Now I find actually temp files useful for debugging, and that is the reason I introduced the data interface.

python/cirq_interface/data_storage_interface.py
95-martin-orion: Discussed offline - solution for now is to omit the `yield`, leaving #174 and #177 unresolved for now. Those can be resolved in a future PR.

PR #198, opened 2019-10-30
Reduce runtime to create container.

s-mandra:


s-mandra:
@alankao64, can you check that this is working? This should speed-up the creation of the container because I've just created a public container with cirq already installed (to avoid that step).

95-martin-orion:
Running `docker-compose up --build` takes **~90s** on my system with this PR, compared to **~700s** on the master branch. This should help significantly in developing using Docker :)

On a separate note: what are the details for the public image? i.e., is it reusable across different OS, similar to normal Docker? Will we need to rebuild it periodically as dependencies are updated?

See #203 for another possible improvement (could be added in parallel to this PR).

s-mandra:
> Running `docker-compose up --build` takes **~90s** on my system with this PR, compared to **~700s** on the master branch. This should help significantly in developing using Docker :)

Great!

> On a separate note: what are the details for the public image? i.e., is it reusable across different OS, similar to normal Docker? Will we need to rebuild it periodically as dependencies are updated?

It's a Docker image I've created and pushed in Docker.hub, so anyone (across OS) can use. It's based on Alpine and install the minimum required to run cirq. I'm planning to create a qFlex Docker image as well.

> See #203 for another possible improvement (could be added in parallel to this PR).

I've tried to do something similar you propose in #203, but it's not easy with Docker. However, I suggest to use a separate PR, so we can improve the life of @alankao64 who needs to run docker many times for testing ;)

alankao64:
> I've tried to do something similar you propose in #203, but it's not easy with Docker. However, I suggest to use a separate PR, so we can improve the life of @alankao64 who needs to run docker many times for testing ;)

Thanks! I'm really looking forward to when this PR gets merged :)

PR #199, opened 2019-10-30
Re-add unit tests for RQC generator

viathor:
Initially added in #74, but dropped in the subsequent move. Made small changes to accommodate changes to `generator.py` that happened since. Specifically, we now read patterns from text files and the test device has a square lattice of four qubits (vs line of three before).

Tested by:

```
$ pytest generator_test.py 
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.7.3, pytest-3.8.2, py-1.8.0, pluggy-0.13.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
plugins: cov-2.5.1, benchmark-3.2.2
collected 4 items                                                                                                                                                                      

generator_test.py ....                                                                                                                                                           [100%]

=============================================================================== 4 passed in 0.04 seconds ===============================================================================
$ mypy generator_test.py 
$
```

### DIFF
--- a/python/circuits/generator_test.py
+++ b/python/circuits/generator_test.py
@@ -1,0 +1,99 @@
+"""Tests for RQC generator.
+
+Invocation:
+    $ pytest generator_test.py
+"""
+
+from typing import Dict, Set, Tuple
+
+import generator
+
+
+def read_patterns(filename):
+    with open(filename, 'r') as f:
+        return eval(f.read())
+
+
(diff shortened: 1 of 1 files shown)


viathor:
Cool! I missed them in that directory. The current tests look overall good to me. They could perhaps have avoided making use of a private property, but that is a minor issue.

PR #200, opened 2019-10-30
Add mypy cache to .gitignore

viathor:


s-mandra:
You may want to consider to add `.mypy_cache/` folder to the variable
```
# Space separated folders in $ROOT_DIR
EXCLUDED_FOLDERS=".env"
```
in `script/check_format.sh` if `.mypy_cache` has some python code (it would avoid to check the format for any `.py` file in `.mypy_cache` folder).

Issue #40, opened 2019-08-30
Migrate away from assert-failures


95-martin-orion:
Assertions aren't the best option for terminating processes, since they don't provide much in the way of logs. We should investigate alternatives (e.g. C++ exceptions) and migrate away from assert() calls.

This also affects our tests, where EXPECT_DEATH doesn't play nice with multithreading.

alankao64:
Resolved by PR #204

PR #204, opened 2019-10-31
Assert removal

alankao64:
Resolves issue #40, and PRs #110 and #189.

### DIFF
--- a/src/contraction_utils.cpp
+++ b/src/contraction_utils.cpp
@@ -1,7 +1,6 @@
 #include "contraction_utils.h"
 
 #include <algorithm>
-#include <cassert>
 #include <fstream>
 #include <iostream>
 #include <regex>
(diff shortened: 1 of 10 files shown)


src/contraction_utils.cpp
```diff
@@ -371,15 +365,14 @@ bool ordering_data_to_contraction_ordering(
     }
   }
   if (!error_msg.empty()) {
-    std::cout << "Parsing failed on line: \"" << line
+    std::cerr << "Parsing failed on line: \"" << line
               << "\" with error: " << error_msg << std::endl;
```
alankao64: Would it make sense to check this to a throw?

src/contraction_utils.cpp
```diff
@@ -509,20 +500,18 @@ bool IsOrderingValid(const std::list<ContractionOperation>& ordering) {
 
   if (std::empty(error_msg)) return true;
 
-  std::cout << error_msg << std::endl;
+  std::cerr << error_msg << std::endl;
```
alankao64: similarly, throw here?

src/contraction_utils.cpp
95-martin-orion: If we did, we would need to update this function to rely on exceptions for error reporting (rather than returning true/false). I think this is the correct direction to go in the future, but it should probably be in a separate PR.

src/contraction_utils.cpp
95-martin-orion: Same as for the other case - probably, but in a future PR.

alankao64:
> Everything looks great! Thanks again for taking this on - I know it's not the most glamorous work, but I really believe that having a clean debug process will encourage new users to adopt and stick with qFlex.

alankao64:
No problem! Glad to help.

Issue #206, opened 2019-11-01
Travis does not install/run YAPF for check_format


95-martin-orion:
From the logs on a recent check:
```
Checking: /home/travis/build/ngnrsaa/qflex/python/qflex_cirq_example.py
scripts/check_format.sh: line 58: yapf3: command not found
Checking: /home/travis/build/ngnrsaa/qflex/python/cirq_interface/fsim_gate.py
scripts/check_format.sh: line 58: yapf3: command not found
{...}
```

PR #209, opened 2019-11-03
Fix check_format.sh to work on Mac OSX and on Travis (YAPF)

s-mandra:


s-mandra:
For consistency, `check_format.sh` is now run inside a docker containers (slightly different versions of `yapf` and `clang-format` give different formats. Also, styles are now saved in `.style.yapf` and `.clang-format` in the root folder.

s-mandra:
Fixes #206

.travis.yml
```diff
@@ -1,15 +1,13 @@
...
   - docker
 
-addons:
-  apt:
-    packages:
-    - clang-format
-
 before_script:
-  - bash scripts/check_format.sh
+  - docker run -ti --rm -v $PWD:/qflex:ro debian:bullseye /bin/bash -c "apt update && apt install -y yapf3 clang-format && bash /qflex/scripts/check_format.sh"
```
95-martin-orion: This check appears to work as intended, but why do we use `debian:bullseye` (instead of e.g. `alpine`)?

scripts/check_format.sh
```diff
@@ -5,6 +5,20 @@
 # be printed as output.
 
 ROOT_DIR="$(realpath $(dirname $0))/../"
+if which yapf3 > /dev/null; then
```
95-martin-orion: I'll defer to @alankao64 for checking this on Mac OSX, as I no longer have easy access to a device running it.

scripts/check_format.sh
```diff
@@ -45,7 +59,7 @@ for filename in $(find_cmd ${ROOT_DIR}/ -type f -iname "*.h" -or -iname "*.cpp")
   filename=$(realpath $(dirname $filename))/$(basename $filename)
   echo "Checking: $filename" >&2
   # ...check if there are any changes required.
-  if clang-format --style=Google --output-replacements-xml "$filename" | grep -q "<replacement "; then
+  if ${CXX_CHECKER} --style=file --output-replacements-xml "$filename" | grep -q "<replacement "; then
```
95-martin-orion: Why did we need to move away from `--style=Google`? Is this related to running in Docker?

Same for the YAPF check.

.clang-format
```diff
@@ -0,0 +1,151 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
95-martin-orion: As I understand it, `BasedOnStyle` suggests that the file contains deviations from that style. Is that the case here, or is this a snapshot of the current Google style? Either way, could you clarify this in the comment?

.style.yapf
```diff
@@ -0,0 +1,332 @@
+[style]
```
95-martin-orion: Same as with `.clang-format` - could you add a comment on whether this is a snapshot of the current Google style?

.travis.yml
s-mandra: `alpine` doesn't have in their repo `yapf` but I can install it through `pip`. However, it makes sense to switch to `alpine`. Thanks!

.clang-format
s-mandra: `.clang-format` should be the format file followed by `clang-format` for the Google style. I dumped it as:
```
clang-format --style google --dump-config > .clang-format
```
Other than that, I'm not familiar with `clang` format.

.style.yapf
s-mandra: Yes, it is by dumping the style as:
```
yapf3 --style google --style-help > .style.yapf
```

.travis.yml
s-mandra: Fixed.

scripts/check_format.sh
s-mandra: Different version of `yapf`/`clang-format` gives different results. Which means that if `yapf` or `clang-format` is just updated, it could fail the build. I've dumped the Google style for for both and put on `.clang-format` and `.style.yapf` (by default, `yapf` looks for this file) in the root folder. Also, I fixed the version of alpine to `3.10.3`. In this way, it is ensured that style remain consistent till we update the corresponding style files.

If you prefer, I can remove this limitation.

95-martin-orion: This seems reasonable.

.style.yapf
95-martin-orion: Could you explain this in a comment in the file? Same for YAPF.

s-mandra: Done, thanks!

scripts/check_format.sh
s-mandra: @alankao64 confirms (after installing `coreutils` using `brew`).

Issue #133, opened 2019-10-21
Support "pip install qflex"


95-martin-orion:
This would simplify the installation process for those using qFlex as a Cirq plugin. Requires [adding qFlex to the Python Package Index](https://packaging.python.org/tutorials/packaging-projects/).

95-martin-orion:
We should time this to line up with our next release (v0.2), making this the final item for the "Post-release fixes" milestone.

alexandrupaler:
Should the `config` folder be shipped with the package? Should we copy it in the cirq interface?

alexandrupaler:
Is it OK if I rename `python/cirq_interface` to `qflexcirq/interface` and place an `__init__.py` in `qflexcirq`?

95-martin-orion:
> Should the `config` folder be shipped with the package? Should we copy it in the cirq interface?

This sounds reasonable. If it's possible to do this programmatically (e.g. in `setup.py`) that would be preferred.

> Is it OK if I rename `python/cirq_interface` to `qflexcirq/interface` and place an `__init__.py` in `qflexcirq`?

I'm in support of both items here, since `qflexcirq` conforms to previous Cirq interface names (e.g. `openfermioncirq`). @s-mandra, will this require changes to the autoconf process?

alexandrupaler:
I hope it is possible to do it programmatically. I will look into this.

alexandrupaler:
The `circuits/generator.py` includes classes which are very similar to Cirq, e.g. `Device`. Should this file be refactored?

alexandrupaler:
> This sounds reasonable. If it's possible to do this programmatically (e.g. in `setup.py`) that would be preferred.

There may be an issue with `example.py` where the configs are used by relative paths

95-martin-orion:
> The `circuits/generator.py` includes classes which are very similar to Cirq, e.g. `Device`. Should this file be refactored?

The namespace distinction (`qflex.Device` vs. `cirq.Device`) is sufficient in theory, but if you see a risk of confusion then e.g. `RQCDevice` might be appropriate.

> There may be an issue with `example.py` where the configs are used by relative paths

Hmm... if there's not a simple solution for this case, I think directly copying a small config for the example shouldn't be a problem.

95-martin-orion:
Once #249 is in, we should add some kind of continuous tests for the install process.

95-martin-orion:
[...]

PR #210, opened 2019-11-04
Pybind can use list of strings as inputs.

s-mandra:


alexandrupaler:
Are strings now the preferred way of sending data to pybind? Three weeks ago I designed the interface around lists of strings, then temp files were added exclusively instead of strings, and this PR is moving once more back to strings exclusively. I am confused with regard to the stability of the interface.  Please see #195 for a discussion about how to solve the bugs introduced by the temp files.

alexandrupaler:
After discussion with Orion, temp file bugs may be solved by using this PR in conjunction with the date interface from #195.

alexandrupaler:
No problem. The intention of the data interface is to have a single place where you open files and automatically close them. I saw that interface as a separation between all of the following : 1) python bindings, 2) file handles and 3) cirq interface classes. We can disable temp files automatically in the data interface, and enable it only when we need some kind of debugging, or even writing QFlex circuits into files to be used without the Cirq interface (in a potential later version of the interface?)

alexandrupaler:
@s-mandra : do you know why Cirq is slow on your cluster? That would be interesting to solve, if possible.

s-mandra:
> @s-mandra : do you know why Cirq is slow on your cluster? That would be interesting to solve, if possible.

No idea .. it can takes 10mins to load it ..

95-martin-orion:
> > @s-mandra : do you know why Cirq is slow on your cluster? That would be interesting to solve, if possible.
> 
> No idea .. it can takes 10mins to load it ..

Just to give context: installing Cirq on my local machine also took 10mins prior to #198, so this isn't exclusive to the cluster.

alexandrupaler:
Do you mean to load by "import cirq" in the script, or to install Cirq from clean. If the later, then Cirq is indeed slow.

95-martin-orion:
> Do you mean to load by "import cirq" in the script, or to install Cirq from clean? If the later, then Cirq is indeed slow.

Specifically installing from clean (i.e. with `docker-compose build`, prior to #198)

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
+    return {};
 
-  return amplitudes;
+  } catch (std::string err_msg) {
+    std::cerr << err_msg << std::endl;
+    return {};
+  } catch (const char *err_msg) {
+    std::cerr << err_msg << std::endl;
+    return {};
+  } catch (...) {
```
95-martin-orion: Instead of catching and discarding unknown error types, we should allow them to be thrown and logged. Even if the error is obscure, getting something specific is more valuable than a generic failure message.

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
-    case 2:
-      std::cerr
-          << "ERROR: either 'ordering_filename' or 'ordering' must be specified"
-          << std::endl;
-      return {};
-  }
+  auto LoadData = [&GetStream](const py::dict &options,
+                               const std::string &argument, auto &data) {
+    if (options.contains(argument.c_str())) {
+      const auto &q = options[argument.c_str()];
```
95-martin-orion: Regarding `const auto &q` - for clarity, could you give a full name to this and similar variables? Currently, it's not immediately clear what this object contains.

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
-  // Check options for circuit
-  switch (options.contains("circuit_filename") + options.contains("circuit")) {
-    case 0:
-    case 2:
-      std::cerr
-          << "ERROR: either 'circuit_filename' or 'circuit' must be specified"
-          << std::endl;
-      return {};
-  }
+  auto GetStream = [](const py::iterable &data, const std::string &caller) {
```
95-martin-orion: These (GetStream, LoadData, LoadState) should be independently-defined functions, with comments describing their behavior. They don't need to appear in `pybind_main.h` though, since we shouldn't need them outside of this file.

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
+    std::stringstream out;
+    for (const auto &line : data)
+      if (py::isinstance<py::str>(line))
+        out << line.cast<std::string>() << std::endl;
+      else
+        throw caller + " must be a list of strings.";
```
95-martin-orion: Can these throw `ERROR_MSG`, or does that cause cross-dependency issues?

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
+                              const std::string &final_state = "") {
+      input.initial_state = initial_state;
+      input.final_state = final_state;
+      return EvaluateCircuit(&input);
+    };
 
-  // Setting initial and final circuit states.
-  if (options.contains("initial_state"))
-    input.initial_state = options["initial_state"].cast<std::string>();
+    // TODO:
```
95-martin-orion: Fill out TODO?

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
-  std::vector<std::pair<std::string, std::complex<double>>> amplitudes =
-      qflex::EvaluateCircuit(&input);
+    if (std::size(initial_states) == 1 and std::size(final_states) == 1)
+      return std::get<1>(amplitudes[0]);
+    else {
+      return {};
+      // return amplitudes;
+    }
+
+    return {};
```
95-martin-orion: Dead code - remove. (Guaranteed to return in the if/else above)

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
+    for (const auto &is : initial_states) {
+      for (const auto &fs : final_states) {
+        amplitudes.push_back({is, run_simulation(is, fs)});
+      }
+    }
 
-  // Evaluating circuit.
-  std::vector<std::pair<std::string, std::complex<double>>> amplitudes =
-      qflex::EvaluateCircuit(&input);
+    if (std::size(initial_states) == 1 and std::size(final_states) == 1)
```
95-martin-orion: Since the inverse of this check already throws an error, this check isn't necessary (we can assume it's true).

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
-    case 0:
-    case 2:
-      std::cerr << "ERROR: either 'grid_filename' or 'grid' must be specified"
-                << std::endl;
-      return {};
-  }
+  auto LoadStates = [](const py::dict &options, const std::string &argument,
+                       const std::size_t num_active_qubits = 0) {
+    std::vector<std::string> states;
+    if (options.contains((argument + "_states").c_str())) {
```
95-martin-orion: Nit: prefer passing the full argument name instead of stitching it together here.

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
+            throw "states must be strings.";
+      } else
+        throw argument + "_states must be a list of strings.";
+    } else if (options.contains((argument + "_state").c_str())) {
+      const auto &q = options[(argument + "_state").c_str()];
+      if (py::isinstance<py::str>(q)) {
+        states.push_back(q.cast<std::string>());
+      } else
+        throw "states must be strings.";
+    } else {
```
95-martin-orion: We should rely on the C++ code to set the initial_state / final_state default values, to ensure that C++ and Python have the same behavior.

src/pybind_main.cpp
```diff
@@ -4,73 +4,114 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
...
-    input.grid.load(options["grid_filename"].cast<std::string>());
-  } else {
-    std::cerr << "ERROR: not yet implemented." << std::endl;
-    return {};
-  }
+    auto run_simulation = [&](const std::string &initial_state = "",
+                              const std::string &final_state = "") {
+      input.initial_state = initial_state;
+      input.final_state = final_state;
+      return EvaluateCircuit(&input);
```
95-martin-orion: This is short enough that it's probably clearer to skip the lambda and move it inside the for loop.

s-mandra:
> > > @s-mandra : do you know why Cirq is slow on your cluster? That would be interesting to solve, if possible.
> > 
> > 
> > No idea .. it can takes 10mins to load it ..
> 
> Just to give context: installing Cirq on my local machine also took 10mins prior to #198, so this isn't exclusive to the cluster.

No, it takes 10mins to import the module :) (not to install it!)

src/pybind_main.cpp
s-mandra: Sure, so do you suggest to not catch them at all if the throw is generic?

src/pybind_main.cpp
s-mandra: Yes, I forgot to implement them, thanks!

src/pybind_main.cpp
s-mandra: Sure, no problems.

src/pybind_main.cpp
s-mandra: Yeah, fair enough. I can remove it.

src/pybind_main.cpp
s-mandra: Sure, I can remove it.

src/pybind_main.cpp
s-mandra: Let me see what I can do here.

src/pybind_main.cpp
s-mandra: Agreed, I can fix this.

src/pybind_main.cpp
s-mandra: Sure, I can do that.

src/pybind_main.cpp
s-mandra: I was keeping it because related to the unfilled `TODO`. But I can remove it and fix it another time.

src/pybind_main.cpp
s-mandra: Unfortunately, I don't find a way to remove the stitching without complicating the code. If possible, I would leave in this way.

src/pybind_main.cpp
s-mandra: Let me know if it works better now.

src/pybind_main.cpp
95-martin-orion: Ah, I guess this is due to the "initial_states" vs "initial_state" distinction. That should be OK, then.

src/pybind_main.cpp
95-martin-orion: > Sure, so do you suggest to not catch them at all if the throw is generic?

That's my recommendation, yes. Unless there's something specific we want to do for these exceptions (e.g. extra log messages in addition to the default message for the error), then not catching the exception is equivalent to `std::rethrow_exception`.

tests/python/simulation_test.py
```diff
@@ -407,28 +388,7 @@
 circuit_filename = mkstemp()
```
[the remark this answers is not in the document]
s-mandra: Fixed.

s-mandra:
@alexandrupaler, let me know your thoughts and thanks a lot for your time!

s-mandra:
> @s-mandra : apologies for reviewing so late. seems ok, but did not find the time to test everything. I trust Orion's opinion. I've been caught with a lot of bureaucracy. Please merge it and I will do my best to integrated this changes in the data interface - or even remove that interface at all. At this point, my main concern is #133

Thanks @alexandrupaler and don't worry, take your time! This PR doesn't require any immediate change to the cirq interface, so no need to rush that now.

PR #215, opened 2019-11-05
Remove qflex-base@docker.

### LINKED ISSUE #214: Automatic qflex@docker images

s-mandra:
Fixes #214. To have automatic `docker` images, I need to remove the initial `qflex-base@docker`.

@alankao64: can you check that everything works? Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ before_script:
   - docker-compose build
 
 script:
-  - docker run -ti --rm qflex-tests:latest
+  - docker run -ti --rm qflex-cxx-tests:latest
   - docker run -ti --rm qflex-py-tests:latest
   - docker run -ti --rm qflex:latest /qflex/config/circuits/bristlecone_48_1-24-1_0.txt /qflex/config/ordering/bristlecone_48.txt /qflex/config/grid/bristlecone_48.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,10 @@ FROM ngnrsaa/cirq-alpine:latest
 
 # Install baseline
 RUN apk update
-RUN apk add git openssh g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
+RUN apk add git g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
 
-# Copy qflex
-COPY ./ /qflex/
+# Clone qflex
+RUN git clone --depth 1 --shallow-submodules https://github.com/ngnrsaa/qflex.git /qflex/
 
 WORKDIR /qflex/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ngnrsaa/cirq-alpine:latest
 
 # Install baseline
 RUN apk update
-RUN apk add g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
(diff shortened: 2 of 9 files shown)


95-martin-orion:
> Fixes #214. To have automatic `docker` images, I need to remove the initial `qflex-base@docker`.

What is the plan for automating this? I assume there's an external tool, but I didn't see anything on the issue description.

s-mandra:
> > Fixes #214. To have automatic `docker` images, I need to remove the initial `qflex-base@docker`.
> 
> What is the plan for automating this? I assume there's an external tool, but I didn't see anything on the issue description.

I've created a special user in github that automatically update a fork of our repo. At that point, `docker` has its own automated tools to keep the docker container updated.

s-mandra:
Also, I'm creating a new `docker` user called `ngnrsaa` so that all the relevant images for this repo are under `ngnrsaa/qflex`, `ngnrsaa/cirq-alpine`, ...

.travis.yml
```diff
@@ -11,6 +11,6 @@ before_script:
   - docker-compose build
 
 script:
-  - docker run -ti --rm qflex-tests:latest
+  - docker run -ti --rm qflex-cxx-tests:latest
```
95-martin-orion: Update docs to match?

Dockerfile
```diff
@@ -1,8 +1,23 @@
...
-RUN apk add g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
+RUN apk add git g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
 
-ENTRYPOINT ["/bin/sh"]
+# Arguments from docker-compose
+ARG OMP_NUM_THREADS
+ARG QFLEX_BRANCH
+
+# Clone qflex
+RUN git clone --depth 1 --branch ${QFLEX_BRANCH:-master} --shallow-submodules ${QFLEX_REPO:-https://github.com/ngnrsaa/qflex.git} /qflex/
```
95-martin-orion: I don't think this is what we want. If someone has changes in their local branch, this will cause Docker runs to ignore those changes.

.travis.yml
s-mandra: Sure, this is still a draft, so I'll eventually do that. Thanks!

Dockerfile
s-mandra: Agreed, however the automated tool to create docker containers doesn't like `COPY` and it only works `git clone`. I'm trying to find different alternatives ..

s-mandra: I was finally able to fix the problem without using `git` but using `COPY`. For some reasons, `docker` hub create a previous snapshots of the submodules and the command `git submodule update --init --recursive` was always failing. The cleanest solution is to not throw an error if `git submodule update` fails.

s-mandra:
Ready to go!

95-martin-orion:
> I've tried it at both `master` and this branch, and I'm seeing the failures below.

For context, `make run-tests` still works correctly for me.

s-mandra:
> > I've tried it at both `master` and this branch, and I'm seeing the failures below.
> 
> For context, `make run-tests` still works correctly for me.

Everything works for me (both `qflex-cxx-tests` and `qflex-py-tests`). Did you clean-up any previous image? That could be an issue that also Alan has found as well. Just to clarify, I've tried:
```
$ docker-compose build
$ docker run -ti --rm qflex-cxx-tests
$ docker run -ti --rm qflex-py-tests
```

95-martin-orion:
> Did you clean-up any previous image?

For the previous tests I had deleted all previous docker containers/images, but I guess there was some cached information that the supposedly-fresh images were making use of.

Deleted and re-cloned git repo and all tests passed.

Issue #216, opened 2019-11-06
Try-catch exceptions


alankao64:
While we currently provide some context for exceptions, methods that can throw exceptions that are hard to trace. Wrapping calls to methods that can throw exceptions in a try-catch block will improve the traceability of errors.

PR #218, opened 2019-11-06
Adding more context to tensor.cpp method calls

alankao64:
Adding try-catch blocks to calls for methods in tensor.cpp for more context when a method fails.
Addresses Issue #216

### DIFF
--- a/src/contraction_utils.cpp
+++ b/src/contraction_utils.cpp
@@ -227,12 +227,12 @@ void ContractionData::ContractGrid(
             try {
               copy_a.project(index, val, tensor_a);
             } catch (std::string err_msg) {
-              throw ERROR_MSG("Failed to perform project(). Error: ", err_msg);
+              throw ERROR_MSG("Failed to call project(). Error: ", err_msg);
             }
             try {
               copy_b.project(index, val, tensor_b);
             } catch (std::string err_msg) {
-              throw ERROR_MSG("Failed to perform project(). Error: ", err_msg);
+              throw ERROR_MSG("Failed to call project(). Error: ", err_msg);
             }
             ContractGrid(ordering, output_index, active_patches);
           }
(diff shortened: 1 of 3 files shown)


src/contraction_utils.cpp
```diff
@@ -102,14 +102,22 @@ ContractionData ContractionData::Initialize(
   long unsigned int max_size = (long unsigned int)pow(bond_dim, max_rank);
 
   // General-purpose scratch space (primarily used for tensor reordering).
-  data.scratch_.push_back(Tensor({""}, {max_size}));
+  try {
+    data.scratch_.push_back(Tensor({""}, {max_size}));
+  } catch (std::string err_msg) {
+    throw ERROR_MSG("Failed to call Tensor(). Error: ", err_msg);
```
alankao64: should there be a new line before the error? `Tensor(). \nError: `

95-martin-orion: I'd recommend trying a couple different things (line breaks, indentation) and going with whatever reads most clearly. One note of caution: we need to make sure that a chain of errors looks different from two separate errors. For example, we don't want to end up with this:
```
Failed to call <methodA>. Error:
Failed to call <methodB>. Error:
<original failure>
```
since it's hard to tell from a single line that it's part of another error.

src/contraction_utils.cpp
```diff
@@ -171,7 +187,11 @@ void ContractionData::ContractGrid(
         Tensor& prev = get_scratch(op.expand.id);
         std::string result_id = result_space(patch_rank_[op.expand.id]);
         Tensor& result = get_scratch(result_id);
-        multiply(prev, next, result, get_scratch(kGeneralSpace).data());
+        try {
+          multiply(prev, next, result, get_scratch(kGeneralSpace).data());
```
95-martin-orion: For `multiply`, `project`, and `reorder`, it would help to include the input tensor(s) in the error message.
(This can be a separate PR if you think that would be cleaner.)

alankao64: I've got another branch where i'm improving error messages, i'll add these fixes to there.

alankao64:
Sample error log:
```
ERROR (some_file.cpp: some_function:123) --> Failed to call another_function() Error:
      [ERROR (another_file.cpp: another_function:234) --> Error message thrown by another_function()]
```

Issue #203, opened 2019-10-31
Add examples for entering a Docker container


95-martin-orion:
Currently, the process for building and testing changes to qFlex with Docker (e.g. when working on a Mac) looks like this:

1. Make desired change in repository.
2. Build and test change with `docker-compose up --build` **(takes X minutes)**
    2a. If all tests pass, you're done.
3. For issues seen in previous test run, apply fixes to the repository. Return to (2)

The bottleneck here is rebuilding the Docker container, even with something like #198 to speed things up. A faster build/test cycle looks like this:

1. Make desired change in repository.
2. Build and test change with `docker-compose up --build` **(takes X minutes)**
    2a. If all tests pass, you're done.
3. Enter the docker container using `${COMMAND_TBD}`
    3a. For issues seen in the previous test run, apply fixes *inside the docker container.*
    3b. Run tests from inside the container using `make run-tests` **(takes XX seconds)**
    3c. If any tests fail, return to (3a).
4. Exit the docker container and replicate all changes to the real repository. Return to (2).

We should either add an example of how to enter the docker container, or update existing examples/Dockerfiles to enter the container (similar to the current rootless-container examples).

s-mandra:
The best approach to fix this would be to pinpoint which files to `COPY` in Dockerfile so that only the relevant layers are updated if some files are missing. Right now, in the most inner layer `qflex`, we copy everything as `COPY ./ /qflex`. Therefore, any change in the folder `./` triggers the update of `qflex` and therefore the recompilation of all layers. This is something I was already thinking about, but it needs some work ..

95-martin-orion:
The examples requested here are largely unnecessary now - the above changes allow `docker-compose up --build` to run just about as fast as a normal `make` would.

PR #220, opened 2019-11-07
Improve dockerfile (depends on: #215)

s-mandra:
This should help issue #203.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,17 +5,29 @@ FROM ngnrsaa/cirq-alpine:latest
 RUN apk update
 RUN apk add git g++ make gsl-dev git autoconf automake python3-dev py3-pybind11 py3-packaging py3-pytest py3-docopt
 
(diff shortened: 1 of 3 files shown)


s-mandra:
This new `Dockerfile`s should allow a better layering. For instance, if only files in `tests/src/` are modified, it only reconstruct the last layer (and it doesn't rebuild the `qflex` layer).

s-mandra:
> I think using a [.dockerignore file](https://docs.docker.com/engine/reference/builder/#dockerignore-file) is the canonical solution to this.

`.dockerignore` would help regarding not copying compiled libraries as `*.o`, but this wouldn't help with a better layering. With the latest modifications, changes to `test/src/` won't affect layers that depends on `src/` only. On the contrary, modifications to `src/` would trigger docker to rebuild layers that depends upon `src/` and `tests/src`.

PR #225, opened 2019-11-08
Improve C++ code.

### LINKED ISSUE #224: Move from int's --> std::size_t where possible.
We got a `C` about `c++` quality because of potential overloads of `int`s. We should move from `int`s to `std::size_t`.

s-mandra:
Should fix #224.


s-mandra:
I've added `-Wall` and `-Wpedantic` to improve the `c++` code.

src/tensor.cpp
```diff
@@ -1038,10 +1043,15 @@ void _generate_binary_reordering_map(
   std::vector<std::size_t> old_counter(num_indices, 0);
   std::size_t po, pn;  // Position of the data, old and new.
   long int i, j;
+
+  // Check that num_indices can be converted to the same type of 'i'
```
95-martin-orion: Is there a reason we can't just use `std::size_t` for i and j instead? (same above)

s-mandra: Wherever I left `long int` instead of `std::size_t` is because some checks are done by assuming that those variables are signed integer. I was planning to eventually fix that as well.

s-mandra: Fixed.

src/tensor.cpp
```diff
@@ -532,9 +545,18 @@ void Tensor::_fast_reorder(std::vector<std::string> new_ordering,
       scratch_copy = NULL;
       return;
     }
+
+    if (Rr == 0) throw ERROR_MSG("Something went wrong.");
+
+    // TODO: This loop has been tested to make sure that extended_Rr is
```
s-mandra: @benjaminvillalonga, is there a way to force this part of the code, so we can test it? I've noticed that large simulations trigger it, but even if I completely comment this loop, the result doesn't change. Thanks!

95-martin-orion:
See comment on issue #224 - and again, apologies for only getting to this after the fact.

s-mandra:
> See comment on issue #224 - and again, apologies for only getting to this after the fact.

As I wrote in #224, here I disagree. this issue is beyond LGTM but clarification of intents. `int`s shouldn't be used freely if `std::size_t` are more appropriate. For instance, when accessing an array, it's more appropriate to use `std::size_t` because it guarantees that `std::size_t` will be large enough to cover the addressable memory. Large part of the conversions where related to that. There are no advantages (in terms of performance) to have `int` rather than `std:size_t`. Moreover, having `std::size_t` helps to early catch errors related to signed conversion that may cause overflow.

Also, I've already made all the conversions :) Everything works properly but there is a loop in `fast_reordering` that has been never checked so far neither with the old implementation nor with my `int`-->`std::size_t` conversion.

s-mandra:
Finally, as a plus, avoiding avoidable conversions using `std::size_t` allowed me to catch all warnings with `-Wall -Wpedantic` to have a more portable code. At the moment, the code is warning free and I'm even using `-Werror` to treat every warning as an error ;) (useful when you want to compile your code using different compilers)

src/read_circuit.cpp
```diff
@@ -211,9 +213,9 @@ std::function<bool(std::vector<int>, std::vector<int>)> order_func(
 
     std::string lpatch = "null";
     std::string rpatch = "null";
-    int lpos = -1;
-    int rpos = -1;
-    int op_num = 0;
+    std::size_t lpos = std::numeric_limits<std::size_t>::max();
```
95-martin-orion: I'd suggest using `std::optional<std::size_t>` to make the "not found" case a little more explicit. Same for other cases where we use `-1` to indicate "not found".

src/tensor.cpp
```diff
@@ -514,10 +524,13 @@ void Tensor::_fast_reorder(std::vector<std::string> new_ordering,
...
-    int Ll = new_binary_ordering.size() - Lr;
-    int Rr = _LOG_2.at(MIN_RIGHT_DIM);
-    int Rl = new_binary_ordering.size() - Rr;
+    if (new_binary_ordering.size() < _LOG_2.at(MAX_RIGHT_DIM))
+      throw ERROR_MSG("Something went wrong.");
+
+    std::size_t Lr = _LOG_2.at(MAX_RIGHT_DIM);
+    std::size_t Ll = new_binary_ordering.size() - Lr;
+    std::size_t Rr = _LOG_2.at(MIN_RIGHT_DIM);
+    // std::size_t Rl = new_binary_ordering.size() - Rr;
```
95-martin-orion: Remove this line?

src/tensor.cpp
```diff
@@ -514,10 +524,13 @@ void Tensor::_fast_reorder(std::vector<std::string> new_ordering,
   // up to L10. Computation times are L4>L5>L6>...>L10. I'll consider
   // all of these cases.
   {
-    int Lr = _LOG_2.at(MAX_RIGHT_DIM);
-    int Ll = new_binary_ordering.size() - Lr;
-    int Rr = _LOG_2.at(MIN_RIGHT_DIM);
-    int Rl = new_binary_ordering.size() - Rr;
+    if (new_binary_ordering.size() < _LOG_2.at(MAX_RIGHT_DIM))
+      throw ERROR_MSG("Something went wrong.");
```
95-martin-orion: Could we give a more specific error message here? Same below.

src/contraction_utils.cpp
```diff
@@ -420,23 +416,26 @@ bool ordering_data_to_contraction_ordering(
...
 
-std::string index_name(const std::vector<int>& p1, const std::vector<int>& p2) {
+std::string index_name(const std::vector<std::size_t>& p1,
+                       const std::vector<std::size_t>& p2) {
   char buffer[64];
   if (p1.size() == 2 && p2.size() == 2) {
     // Two-qubit contraction.
-    int len = snprintf(buffer, sizeof(buffer), "(%d,%d),(%d,%d)", p1[0], p1[1],
-                       p2[0], p2[1]);
+    std::size_t len = snprintf(buffer, sizeof(buffer), "(%ld,%ld),(%ld,%ld)",
```
95-martin-orion: For `size_t`, use `%zd`: [docs](http://www.cplusplus.com/reference/cstdio/printf/)

Same elsewhere in PR.

src/read_circuit.cpp
s-mandra: Fixed (and removed the `-1` for not found).

src/contraction_utils.cpp
```diff
@@ -364,26 +380,19 @@ void ordering_data_to_contraction_ordering(
...
-      if (index < 0) {
-        error_msg = "Index cannot be negative.";
-        break;
-      }
-      if (index >= input.grid.I * input.grid.J) {
-        error_msg = "Index must be within grid boundaries.";
-        break;
-      }
-      std::vector<int> position = get_qubit_coords(index, input.grid.J);
+      if (not check_index(index, error_msg)) break;
```
s-mandra: Perhaps, it would be better to have `check_index` just throw in case of an error.

src/contraction_utils.cpp
```diff
@@ -468,23 +465,26 @@ void ordering_data_to_contraction_ordering(
   }
 }
 
-std::string index_name(const std::vector<int>& p1, const std::vector<int>& p2) {
+std::string index_name(const std::vector<std::size_t>& p1,
+                       const std::vector<std::size_t>& p2) {
   char buffer[64];
```
s-mandra: We should remove `buffer` to avoid buffer overflow.

src/evaluate_circuit.cpp
```diff
@@ -236,7 +223,7 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
 
     // Contract 3D grid onto 2D grid of tensors, as usual.
     flatten_grid_of_tensors(tensor_grid_3D, tensor_grid, final_qubits,
-                            input->grid.qubits_off, ordering, scratch_2D);
+                            input->grid.qubits_off, ordering, &scratch_2D[0]);
```
95-martin-orion: Prefer `scratch_2D.data()` for a pointer to the head of the array.

s-mandra: Agreed, done!

src/read_circuit.h
```diff
@@ -50,30 +50,32 @@ namespace qflex {
 const double _SQRT_2 = 1.41421356237309504880168872;
```
s-mandra: Probably, it would be better to use <cmath> constants here.

s-mandra:
I'm happy with this PR. @95-martin-orion, let me know if you spot something else. Thanks for your help!

alankao64:
Confirmed these changes do not break any tests in master or in other existing branches.

PR #227, opened 2019-11-08
Fix circuit.load()

alankao64:
Bug in circuit.load().

In read_circuit_test.cpp, ReadCircuitTest, CircuitReferencingInactiveQubits:
`circuit.load(std::stringstream(kBad_Gate))` is not updating the circuit, so `circuit_data_to_tensor_network()` is getting called on `kBadTGate` for all tests.

### DIFF
--- a/src/read_circuit.cpp
+++ b/src/read_circuit.cpp
@@ -366,6 +366,15 @@ void circuit_data_to_tensor_network(
     idx += 1;
   }
 
+  // Check if off contains valid qubits
+  if (off.has_value())
+    for(const auto &w: off.value()) {
+      const auto &x = w[0];
+      const auto &y = w[1];
+      if (x < 0 or x >= I or y < 0 or y >= J)
+        throw ERROR_MSG("Off qubit '(", x, ", ", y, ")' is outside the grid.");
+    }
+
   std::unordered_set<int> used_qubits;
   std::size_t last_cycle{0};
   for (auto gate : circuit.gates) {
--- a/src/read_circuit.cpp
+++ b/src/read_circuit.cpp
@@ -396,8 +396,9 @@ void circuit_data_to_tensor_network(
       // Check that position is an active qubit
       bool qubit_off = find_grid_coord_in_list(off, i_j_1[0], i_j_1[1]);
       if (qubit_off)
-        throw ERROR_MSG("Qubit '", q1, "' in gate '", gate.cycle, " ", gate.name, " ",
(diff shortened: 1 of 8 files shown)


s-mandra:
@alankao64, I'm running `make -j8 && make -j8 run-tests` and everything is fine. Is this what I would expect? If not, could you adjust the test to trigger an error, so I can work from that? Thanks!

alankao64:
> @alankao64, I'm running `make -j8 && make -j8 run-tests` and everything is fine. Is this what I would expect? If not, could you adjust the test to trigger an error, so I can work from that? Thanks!

Everything should pass, the reason I found this bug is because we wanted to add the gate into the error message, and then I found that circuit.load() wasn't actually updating circuit. I'll adjust the test to fail.

alankao64:
@s-mandra the test should be failing now!

s-mandra:
Everything works now. Thanks for catching this @alankao64!

tests/src/read_circuit_test.cpp
```diff
@@ -103,9 +103,7 @@ constexpr char kBadTGate[] = R"(1
 constexpr char kBadCzGate[] = R"(1
 1 cz 0 1)";
 
-constexpr char kBadCxGate[] = R"(1
-1 cx 0 1)";
-
+// Broken test
```
95-martin-orion: Is this test still broken?

src/read_circuit.cpp
```diff
@@ -395,7 +405,9 @@ void circuit_data_to_tensor_network(
       // Check that position is an active qubit
       bool qubit_off = find_grid_coord_in_list(off, i_j_1[0], i_j_1[1]);
       if (qubit_off)
-        throw ERROR_MSG("Qubit ", q1, " must correspond to an active qubit.");
+        throw ERROR_MSG("Qubit '", q1, "' in gate '", gate.cycle, " ",
```
95-martin-orion: If possible, these should use a method in QflexGate to print the gate - maybe store the exact string for each gate when loading the circuit? The log here will skip `gate.params` if it has any.

Same for 2-qubit gates.

tests/src/read_circuit_test.cpp
s-mandra: No, everything should work. Let me remove the comment.

src/read_circuit.cpp
s-mandra: Let me take a look into it.

s-mandra: Fixed.

PR #228, opened 2019-11-09
Fix string catch.

s-mandra:
`catch (std::string)` --> `catch (const std::string&)`


s-mandra:
@alankao64, for future reference, try to catch by reference when possible, for instance:
```
catch (const std::string& err_msg)
```
instead of
```
catch (std::string err_msg)
```
Thanks!

PR #229, opened 2019-11-09
Fix index_value (depends on: #228)

s-mandra:


src/tensor.cpp
```diff
@@ -237,7 +237,7 @@ void Tensor::project(std::string index, size_t index_value,
     throw ERROR_MSG("Index: '", index, "' has to be equal to indices[0]: '",
                     _indices[0], "'.");
   }
-  if (index_value < 0 || index_value >= _dimensions[0]) {
+  if (index_value >= _dimensions[0]) {
```
s-mandra: Fixed this test given that `index_value` is `std::size_t` and therefore never negative. This is the only fix.

s-mandra:
> Thanks for the fix! It would be nice (but not strictly necessary) if we could get LGTM to run automatically before merging each PR.

Agreed. My understanding is that you need to install their app in order to do that ..

95-martin-orion:
> Agreed. My understanding is that you need to install their app in order to do that ..

Ah :( In that case, the current nightly runs should be sufficient. Thanks for checking!

PR #231, opened 2019-11-11
Try-catch for evaluate_circuit.cpp and contraction_utils.cpp

alankao64:
Added try-catch for calls to methods in evaluate_circuit.cpp and contration_utils.cpp that can throw exceptions.


src/evaluate_circuit.cpp
```diff
@@ -60,7 +60,11 @@ std::string get_output_states(const QflexInput* input,
     if (op.op_type != ContractionOperation::CUT) continue;
     // Any qubit with a terminal cut is in the final region.
     if (op.cut.tensors.size() != 1) continue;
-    const int pos = find_output_pos(input, op.cut.tensors[0]);
+    try {
+      const int pos = find_output_pos(input, op.cut.tensors[0]);
+    } catch (const std::string& err_msg) {
+      throw ERROR_MSG("Failed to call find_output_pos(). Error:\n\t[", err_msg, "]");
+    }
```
alankao64: moving the declaration of `const int pos` inside the try catch is causing errors at line 70

alankao64:
Still in draft, don't review just yet.

src/evaluate_circuit.cpp
alankao64: Also is causing problems at a lot of other places

alankao64:
changed `const std::string` to `std::string`
moved variable declarations outside try-catch blocks

src/contraction_utils.cpp
```diff
@@ -237,7 +249,12 @@ void ContractionData::ContractGrid(
               throw ERROR_MSG("Failed to call project(). Error:\n\t[", err_msg,
                               "]");
             }
-            ContractGrid(ordering, output_index, active_patches);
+            try {
+              ContractGrid(ordering, output_index, active_patches);
+            } catch (const std::string& err_msg) {
+              throw ERROR_MSG("Failed to call ContractGrid(). Error:\n\t[",
```
95-martin-orion: Re-throwing here will generate log spam - each layer of recursion will append another (redundant) copy of this message. The try-catch block can be removed.

src/contraction_utils.cpp
```diff
@@ -251,7 +268,12 @@ void ContractionData::ContractGrid(
               throw ERROR_MSG("Failed to call project(). Error:\n\t[", err_msg,
                               "]");
             }
-            ContractGrid(ordering, output_index, active_patches);
+            try {
+              ContractGrid(ordering, output_index, active_patches);
```
95-martin-orion: Same here.

src/evaluate_circuit.cpp
```diff
@@ -60,7 +60,13 @@ std::string get_output_states(const QflexInput* input,
     if (op.op_type != ContractionOperation::CUT) continue;
     // Any qubit with a terminal cut is in the final region.
     if (op.cut.tensors.size() != 1) continue;
-    const int pos = find_output_pos(input, op.cut.tensors[0]);
+    std::size_t pos;
```
95-martin-orion: (Pre-existing issue) We're reusing the variable name `pos` in this method. It works, but it's likely to cause issues in the future - could you change it to something more specific?

src/evaluate_circuit.cpp
```diff
@@ -60,7 +60,13 @@ std::string get_output_states(const QflexInput* input,
...
+    try {
+      pos = find_output_pos(input, op.cut.tensors[0]);
+    } catch (const std::string& err_msg) {
+      throw ERROR_MSG("Failed to call find_output_pos(). Error:\n\t[", err_msg,
+                      "]");
+    }
```
95-martin-orion: Could you add a comment after this line noting that `pos` is guaranteed to be initialized if no error is caught?

Same thing for similar "initialize-then-try" patterns in this PR.

src/evaluate_circuit.cpp
alankao64: renamed to output_pos

src/contraction_utils.cpp
```diff
@@ -209,74 +208,63 @@ void ContractionData::ContractGrid(
...
-              throw ERROR_MSG("Failed to call ContractGrid(). Error:\n\t[",
-                              err_msg, "]");
+              output_index++;
             }
-            output_index++;
+            tensor_a = copy_a;
           }
-          tensor_a = copy_a;
+        } catch (const std::string& err_msg) {
+          throw ERROR_MSG("Failed to during CUT. Error:\n\t[", err_msg, "]");
```
95-martin-orion: `Failed to` (???) `during CUT` -> `Failed during CUT`? Or fill in the (???).

Also note that this will trigger *in addition* to the interior try-catch blocks - for example:
```
Failed during CUT. Error:
    Failed to call project(). Error:
    <error message>
```

src/contraction_utils.cpp
```diff
@@ -65,20 +65,19 @@ ContractionData ContractionData::Initialize(
         break;
       }
       case ContractionOperation::CUT: {
-        std::string cut_index;
         try {
-          cut_index = index_name(op.cut.tensors);
+          const std::string cut_index = index_name(op.cut.tensors);
```
95-martin-orion: Just to be safe, could you run a couple of simulations both with and without this PR, and check the runtime for each? (e.g. bristlecone_48 and sycamore_53) I doubt this will significantly impact performance, but I'd rather be certain.

src/contraction_utils.cpp
```diff
@@ -209,74 +208,63 @@ void ContractionData::ContractGrid(
...
+                throw ERROR_MSG("Failed to call project(). Error:\n\t[",
+                                err_msg, "]");
+              }
+              try {
+                copy_b.project(index, val, tensor_b);
+              } catch (const std::string& err_msg) {
+                throw ERROR_MSG("Failed to call project(). Error:\n\t[",
+                                err_msg, "]");
+              }
+              ContractGrid(ordering, output_index, active_patches);
```
95-martin-orion: So this is awkward. An error in this `ContractGrid` call will be caught by the outer try-catch, which puts us in the same log-spam situation as before. I think it's probably best just to treat `index_name` as "fundamental".

src/contraction_utils.cpp
alankao64: With and without PR:
- bristlecone_48: ~5.6 seconds
- sycamore_53: ~0.2 seconds

Seems like there's no change.

s-mandra:
For some reasons, `evaluate_circuit_test.cpp` doesn't like this ordering anymore:
```
cut () 1 3
expand a 1
expand a 0
expand a 2
cut () 5
expand b 5
expand b 3
merge a b
```
any thoughts of the reason why?

s-mandra:
> For some reasons, `evaluate_circuit_test.cpp` doesn't like this ordering anymore:
> 
> ```
> cut () 1 3
> expand a 1
> expand a 0
> expand a 2
> cut () 5
> expand b 5
> expand b 3
> merge a b
> ```
> 
> any thoughts of the reason why?

```
ERROR (contraction_utils.cpp: ordering_data_to_contraction_ordering:471) --> 
Index (0,1),(1,1) is cut multiple times.
Tensor at (0,1) is added to non-empty patch a after a cut.
Tensor at (0,1) is added to non-empty patch a after a cut.
Tensor (0,1) is contracted multiple times.
Tensor at (0,0) is added to non-empty patch a after a cut.
Tensor at (0,0) is added to non-empty patch a after a cut.
Tensor (0,0) is contracted multiple times.
Tensor at (1,0) is added to non-empty patch a after a cut.
Tensor at (1,0) is added to non-empty patch a after a cut.
Tensor (1,0) is contracted multiple times.
Index (2,1),(o) is cut multiple times.
Tensor at (2,1) is added to non-empty patch b after a cut.
Tensor (2,1) is contracted multiple times.
Tensor at (1,1) is added to non-empty patch b after a cut.
Tensor (1,1) is contracted multiple times.
Patch a is merged multiple times.
Patch a is merged into non-empty patch b after a cut.
```

95-martin-orion:
> For some reasons, `evaluate_circuit_test.cpp` doesn't like this ordering anymore:
> {...ordering...}
> any thoughts of the reason why?

Pretty sure this is caused by the duplicate `ordering_data_to_contraction_ordering` call in evaluate_circuit.cpp:135-143.

s-mandra:
> > For some reasons, `evaluate_circuit_test.cpp` doesn't like this ordering anymore:
> > {...ordering...}
> > any thoughts of the reason why?
> 
> Pretty sure this is caused by the duplicate `ordering_data_to_contraction_ordering` call in evaluate_circuit.cpp:135-143.

Thanks! I was wondering if I made a mistake once I merged `master` to `try-catch` (other examples were indeed broken!)

s-mandra:
> > For some reasons, `evaluate_circuit_test.cpp` doesn't like this ordering anymore:
> > {...ordering...}
> > any thoughts of the reason why?
> 
> Pretty sure this is caused by the duplicate `ordering_data_to_contraction_ordering` call in evaluate_circuit.cpp:135-143.

Everything should work now!

s-mandra:
> See comments - based on the scope of `index_name` return values, we probably need to treat it as a "fundamental" function. @s-mandra, does this seem reasonable?

Just had a discussion with @alankao64. Given that `CUT` is the only function that calls `index_name`, perhaps it makes sense to have `index_name` to just return an empty string in case of error and let `CUT` handle the error. @95-martin-orion, what is your opinion? Thanks!

95-martin-orion:
> {...} Given that `CUT` is the only function that calls `index_name`, perhaps it makes sense to have `index_name` to just return an empty string in case of error and let `CUT` handle the error.

`read_circuit.cpp` also makes calls to `index_name`. Since we probably want a similar error message at both call sites, I'd prefer throwing the error in `index_name` and leaving it uncaught in `ContractionData::ContractGrid`. The error will percolate up and be caught in the public `ContractGrid` method (line ~600), which should provide enough detail for users.

s-mandra:
> > {...} Given that `CUT` is the only function that calls `index_name`, perhaps it makes sense to have `index_name` to just return an empty string in case of error and let `CUT` handle the error.
> 
> `read_circuit.cpp` also makes calls to `index_name`. Since we probably want a similar error message at both call sites, I'd prefer throwing the error in `index_name` and leaving it uncaught in `ContractionData::ContractGrid`. The error will percolate up and be caught in the public `ContractGrid` method (line ~600), which should provide enough detail for users.

Sounds good to me!

src/contraction_utils.cpp
alankao64: Removed this catch.

src/contraction_utils.cpp
```diff
@@ -463,6 +459,8 @@ bool ordering_data_to_contraction_ordering(
   if (!error_msg.empty()) {
     std::cerr << "Parsing failed on line: \"" << line
               << "\" with error: " << error_msg << std::endl;
+    // throw ERROR_MSG("Parsing failed on line: '", line, "' with error: ",
```
95-martin-orion: Why do we return `pair<bool, string>` from this function? I feel like returning `void` and `throw`ing any errors is cleaner - the caller can `try-catch` to discard the error if it shouldn't cause a failure.

@s-mandra, your thoughts?

95-martin-orion: Tangentially related: we should either delete the log line or delete the comment here, depending on the decision for this.

s-mandra: > Why do we return `pair<bool, string>` from this function? I feel like returning `void` and `throw`ing any errors is cleaner - the caller can `try-catch` to discard the error if it shouldn't cause a failure.
> 
> @s-mandra, your thoughts?

Mhh .. let me say that it would be strange to have a function called `IsOrderingValid` which returns `void` and `throw` in case it isn't. I also agree that it would be cleaner to avoid a pair `{bool, string}` to catch this error. I'm ok with either ways, but we need to adjust few tests because they rely on the `{bool, string}` pair.

s-mandra: I'm actually following @95-martin-orion suggestion.

95-martin-orion: > ...it would be strange to have a function called `IsOrderingValid` which returns `void` and `throw` in case it isn't.

Maybe rename to `ValidateOrdering` or similar?

s-mandra: > > ...it would be strange to have a function called `IsOrderingValid` which returns `void` and `throw` in case it isn't.
> 
> Maybe rename to `ValidateOrdering` or similar?

Yeah .. `ValidateOrdering` sounds better than `IsOrderingValid`. Let me change it.

s-mandra:
I think we

s-mandra:
@alankao64, remember to catch reference to constant strings as `const std::string& msg`, not strings.

alankao64:
> @alankao64, remember to catch reference to constant strings as `const std::string& msg`, not strings.

Got it, sorry about that!

s-mandra:
> Looks good to me! In a follow-up PR, let's update `contraction_utils_test` to use more specific exception testing in place of `EXPECT_ANY_THROW`.

Fixed.

s-mandra:
At the moment, `ordering_data_to_contraction_ordering` doesn't properly rethrow errors and some useful info are just printed to `std::cerr`. This requires some extra changes in `evaluate_circuit.cpp`, so I'll open a new PR for this.

s-mandra:
> At the moment, `ordering_data_to_contraction_ordering` doesn't properly rethrow errors and some useful info are just printed to `std::cerr`. This requires some extra changes in `evaluate_circuit.cpp`, so I'll open a new PR for this.

Fixed in #248.

PR #232, opened 2019-11-12
Circuit.cpp and grid.cpp tests

alankao64:
Adding more tests.

### DIFF
--- a/src/grid.cpp
+++ b/src/grid.cpp
@@ -43,7 +43,7 @@ void QflexGrid::load(const std::string& filename) {
   if (auto in = std::ifstream(filename); in.good())
     this->load(in);
   else
-    throw ERROR_MSG("Cannot open grid file: ") + filename;
+    throw ERROR_MSG("Cannot open grid file: ", filename, ".");
 }
 
 }  // namespace qflex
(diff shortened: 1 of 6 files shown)


tests/src/circuit_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+4 cz 2 4
+9 cz 0 2
+9 cz 1 3
+9 h 1)";
+
+// Testing bad circuits
+TEST(CircuitExceptionTest, BadCircuits) {
+  QflexCircuit circuit;
+
+  // First line isn't the number of active qubits
```
95-martin-orion: Just for readability, could you copy these comments to the corresponding `kBadCircuit` definitions above?

src/grid.cpp
```diff
@@ -26,7 +26,7 @@ void QflexGrid::load(std::istream&& istream) {
 
       // Get number of columns
       if (J != 0 and J != std::size(line))
-        throw std::string("Grid size is inconsistent");
+        throw ERROR_MSG("Grid size is inconsistent.");
```
95-martin-orion: Could you add J and the line size to this log? (Line number might also be helpful.)

Same for the corresponding test.

PR #233, opened 2019-11-13
Add verbose option

s-mandra:
These changes would allow the user to decide the level of logs to print while running.

### DIFF
--- a/src/global.h
+++ b/src/global.h
@@ -1,0 +1,6 @@
+#ifndef GLOBAL__H
+#define GLOBAL__H
+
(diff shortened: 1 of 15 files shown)


src/global.h
```diff
@@ -0,0 +1,6 @@
+#ifndef GLOBAL__H
+#define GLOBAL__H
+
+inline int VERBOSE = 0;
```
95-martin-orion: Actually, can we wrap these in a namespace? `qflex::global` would be my preference.

s-mandra: Sounds good.

Issue #141, opened 2019-10-22
Large circuits cause out-of-memory errors


alexandrupaler:
The cast to int returns 0, and the program fails with 

MemoryError: std::bad_array_new_length

when super_dim==32, for example

```
libc.so.6!__GI_raise(int sig) (/build/glibc-OTsEL5/glibc-2.27/sysdeps/unix/sysv/linux/raise.c:51)
libc.so.6!__GI_abort() (/build/glibc-OTsEL5/glibc-2.27/stdlib/abort.c:79)
libstdc++.so.6![Unknown/Just-In-Time compiled code] (Unknown Source:0)
libstdc++.so.6!std::terminate() (Unknown Source:0)
libstdc++.so.6!__cxa_throw (Unknown Source:0)
libstdc++.so.6!__cxa_throw_bad_array_new_length (Unknown Source:0)
qflex::EvaluateCircuit[abi:cxx11](qflex::QflexInput*)(qflex::QflexInput * input) (/home/alexandru/github/qflex/src/evaluate_circuit.cpp:186)
main(int argc, char ** argv) (/home/alexandru/github/qflex/src/main.cpp:96)
```

https://github.com/ngnrsaa/qflex/blob/c7038b0e06d2f9312a3b5db41dcd9a244f276220/src/evaluate_circuit.cpp#L186

s-mandra:
@95-martin-orion, should we close this issue?

95-martin-orion:
> @95-martin-orion, should we close this issue?

Seems reasonable. The memory flag should be sufficient guard against this issue in the future.

PR #236, opened 2019-11-14
Add --memory flag to prevent OOM failures.

95-martin-orion:
Determine the required space before attempting to allocate it, and cancel the simulation if required space exceeds a user-specified limit (default 1GB). Try it out:
```
$ src/qflex.x config/circuits/bristlecone_48_1-24-1_0.txt config/ordering/bristlecone_48.txt config/grid/bristlecone_48.txt -m 20000000

# should return:
ERROR (contraction_utils.cpp: Initialize:145) --> Required space (106.799 MB) exceeds memory limit (19.0735 MB). Cancelling simulation.
```

Also replaced `long unsigned int` with `uint64_t`, since on some systems `long` will only use 32 bits.

This partially addresses #141, but does not fully resolve it (tensors larger than rank 32 will still cause issues until #225 is submitted).

### DIFF
--- a/src/contraction_utils.cpp
+++ b/src/contraction_utils.cpp
@@ -11,6 +11,22 @@
 
 namespace qflex {
 
+namespace {
+
+// Convert 'memory' bytes into a more readable string.
+std::string readable_memory_string(double memory) {
+  std::string suffix[] = {" B", " kB", " MB", " GB"};
+  std::size_t scale = 0;
+  double size_prefix = memory;
+  while (size_prefix >= (1 << 10)) {
+    ++scale;
+    size_prefix /= (1 << 10);
+  }
+  return concat(size_prefix, suffix[scale]);
+}
+
+}  // namespace
+
 // ContractionData methods
 
 ContractionData ContractionData::Initialize(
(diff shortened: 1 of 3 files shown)


src/contraction_utils.cpp
```diff
@@ -92,14 +108,56 @@ ContractionData ContractionData::Initialize(
     }
   }
 
-  long allocated_space = 0;
+  // Calculate the necessary space before actually allocating memory.
+
+  // If the --memory flag is set properly, this should prevent OOM errors.
+  uint64_t allocated_space = 0;
```
s-mandra: For consistency, we should use `std::size_t`.

s-mandra: Similarly elsewhere.

s-mandra: But #225 is not yet merged, so I'm ok to keep `uint64_t` for now.

s-mandra:
Looks good to me! I'll fix `uint64_t` --> `std::size_t` once we merge #225.

src/contraction_utils.cpp
95-martin-orion: Done.

95-martin-orion:
There were a couple of unneeded `allocated_space += {...}` lines, so I went ahead and changed to `std::size_t` along with that change.

PR #237, opened 2019-11-15
Fix check-format.py to properly exclude modules in Mac OSX.

s-mandra:


scripts/check_format.sh
```diff
@@ -30,23 +30,23 @@ function find_cmd() {
   shift
 
   # Get modules
-  modules=$(cat ${ROOT_DIR}/.gitmodules 2>/dev/null | grep path | sed 's/[[:space:]]*//g' | awk -F "=" -v root_dir=${ROOT_DIR} '{ print root_dir"/"$2 }' | tr '\n' '|')
+  modules=$(cat ${ROOT_DIR}/.gitmodules 2>/dev/null | grep path | sed 's/[[:space:]]*//g' | awk -F "=" -v root_dir=${ROOT_DIR} '{ print root_dir"/"$2 }' | xargs realpath | tr '\n' '|')
```
95-martin-orion: Does this work for modules with spaces in their name? It looks like `sed 's/[[:space:]]*//g'` deletes all spaces in the filename, which wouldn't match the original name.

Not too big of an issue for now, since none of our modules have spaces.

s-mandra: Good point. Let me think about it.

s-mandra: Fixed. Actually, you don't really need to call `sed`, so I removed it.

alankao64:
> Leaving Mac OSX checks to @alankao64.

Seems like check_format.sh now correctly ignores .cpp files in .gitmodules, but isn't catching incorrectly formatted python files in general.

s-mandra:
> > Leaving Mac OSX checks to @alankao64.
> 
> Seems like check_format.sh now correctly ignores .cpp files in .gitmodules, but isn't catching incorrectly formatted python files in general.

I made different improvement to the script. Right now it should:

- exclude all ignored files (include files in ignored folders)
- exclude files in submodules
- handle files with white spaces

@alankao64, would you mind to test this script to make sure that it works on Mac OSX as well? Thanks!

alankao64:
Not sure if its just my Mac, but incorrectly formatted python files aren't being caught properly.

s-mandra:
Strange .. it works properly on my Mac OS X workstation. Let me take a closer look.

s-mandra:
There was an error that prevented Travis to properly check the format. Now everything works. However, I've noticed that `yapf` version is very important: indeed, different version still gives different formatting even if the style is kept fixed. I'm trying to address this problem by standardizing the required version.

alexandrupaler:
yapf is a python module, and I tried to standardize its version in requirements.txt

s-mandra:
> yapf is a python module, and I tried to standardize its version in requirements.txt

Thanks! There is also a python module for `clang-format`, but it cannot install it on alpine ..

s-mandra:
> > yapf is a python module, and I tried to standardize its version in requirements.txt
> 
> Thanks! There is also a python module for `clang-format`, but it cannot install it on alpine ..

Fixed. Now `.travis.yml` uses `requirements.txt` to determine the right version to install (through `pip`).

s-mandra:
`check_format.sh` now requires `yapf` and `clang-format` to be the exact same version as reported in `requirements.txt`.

PR #238, opened 2019-11-15
Only allocate as much space as needed for 2D grid.

### LINKED ISSUE #162: Allocate reading scratch space based on contraction, not on depth.

95-martin-orion:
Fixes #162. `scratch_2D` and `scratch_3D` are now destroyed when they go out of scope; no explicit `delete` is required.

### DIFF
--- a/src/evaluate_circuit.cpp
+++ b/src/evaluate_circuit.cpp
@@ -12,6 +12,8 @@
  * @licence: Apache License, Version 2.0
  */
 
+#include <set>
+
 #include "evaluate_circuit.h"
 
 namespace qflex {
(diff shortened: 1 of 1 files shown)


src/evaluate_circuit.cpp
```diff
@@ -148,61 +150,75 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
...
+          }
+        }
+        if (indices.size() > max_rank) {
+          max_rank = indices.size();
+        }
+      }
+    }
+    // Scratch space for contracting 3D to 2D grid. This must have enough space
+    // to hold the largest single-qubit tensor in the 2D grid.
+    s_type scratch_2D[static_cast<std::size_t>(pow(2, max_rank))];
```
benjaminvillalonga: I think this doesn't take into account that some indexes might have dim > 2. For example, the virtual index of an fSim gate. A fix could involve using get_index_to_dimension() from Tensor.

benjaminvillalonga: Everything else looks good to me.

95-martin-orion: Updated to use `get_index_to_dimension()`.

PR #241, opened 2019-11-26
Improve reordering tests

alankao64:
Adding more tests cases for reordering of tensor indices.


tests/src/tensor_test.cpp
```diff
@@ -11,6 +11,36 @@ namespace {
 using ::testing::Eq;
 using ::testing::Pointwise;
 
+// Simple tensor permutation of axes
+template <typename array_type, typename idx1_type, typename idx2_type>
+array_type simple_reordering(const array_type &array, const idx1_type &idx1,
```
95-martin-orion: Could you add comments explaining these arguments and their types?

tests/src/tensor_test.cpp
```diff
@@ -164,6 +194,169 @@ TEST(TensorTest, IndexReordering) {
   }
 }
 
+// Tests a reordering with a single right move.
```
95-martin-orion: For clarity, it might help to include a visualization of each test, showing the left, middle and right sections. e.g., for this one:
```
<empty> | abcde | fghij
-> <empty> | jegcf | bhida
```

tests/src/tensor_test.cpp
```diff
@@ -164,6 +194,169 @@ TEST(TensorTest, IndexReordering) {
...
+
+  // Check Tensor data.
+  data = simple_reordering(data, indices, expected_indices);
+
+  for (std::size_t i = 0; i < std::size(data); ++i) {
+    ASSERT_EQ(data[i], tensor.data()[i]);
+  }
+}
+
+TEST(TensorTest, SlowLeftTwelveIndicesReordering) {
```
95-martin-orion: Does this cover a different code path than the `LeftTwelveIndicesReordering` test above? If not, we can consolidate to a single test.

tests/src/tensor_test.cpp
```diff
@@ -164,6 +194,169 @@ TEST(TensorTest, IndexReordering) {
...
+    data.push_back(std::complex<float>(i, 0));
+  }
+  Tensor tensor(indices, dimensions, data);
+  std::vector<std::string> expected_indices = {"a", "b", "l", "d", "e", "f",
+                                               "h", "g", "i", "j", "k", "c"};
+  std::array<std::complex<float>, 4096> scratch;
+  try {
+    tensor.reorder(expected_indices, scratch.data());
+  } catch (std::string msg) {
+    std::cout << msg << std::endl;
```
95-martin-orion: Should the test fail if `tensor.reorder` fails? Same elsewhere in the file.

alankao64: You're right, it definitely should.

tests/src/tensor_test.cpp
alankao64: The code also checks to see if it can perform the reordering up to the leftmost 8 indices. This case has a reordering using the 8th index, which is in the rightmost group, as opposed to the other test that keeps the reordering in the two left groups.

tests/src/tensor_test.cpp
s-mandra: Done! Let me know if it works for you or if you think something more detailed would be better.

tests/src/tensor_test.cpp
95-martin-orion: Interesting...the visual representation of this contradicts my understanding of how left-reorder works. How is a left-reorder able to manipulate an index in the rightmost region?

tests/src/tensor_test.cpp
```diff
@@ -304,7 +344,9 @@ TEST(TensorTest, SlowLeftTwelveIndicesReordering) {
   }
 }
 
-// Tests a reordering with a left and a right move.
+// Tests a reordering needing a left and a right move.
+// ab | cdefg | hijkl
+// -> cd | hijkl | efgab
```
95-martin-orion: Could you include each step of the reordering in this visual?

tests/src/tensor_test.cpp
```diff
@@ -331,7 +375,9 @@ TEST(TensorTest, LeftRightIndexReordering) {
   }
 }
 
-// Tests a worse case index reordering.
+// Tests a worse case index reordering needing a left, right, and left move.
+// ab | cdefg | hijkl
+// -> kc | aldgf | hbeij
```
95-martin-orion: Same here - please show each step of the reorder.

alankao64:
> Interesting...the visual representation of this contradicts my understanding of how left-reorder works. How is a left-reorder able to manipulate an index in the rightmost region?

From my understanding of what @benjaminvillalonga said, the left-reorder will check up to 8 indices from the left and if it can complete the reordering by doing a left move with those indices then it will do so.

95-martin-orion:
> ...the left-reorder will check up to 8 indices from the left and if it can complete the reordering by doing a left move with those indices then it will do so.

Huh. In that case, I guess leave the test as-is, but update the comment with this explanation + updated section boundaries.

tests/src/tensor_test.cpp
```diff
@@ -311,9 +311,11 @@ TEST(TensorTest, LeftTwelveIndicesReordering) {
   }
 }
 
-// Tests a reordering needing a single left move, checks up to the 8 indices.
-// ab | cdefg | hijkl
-// -> hb | cdefg | aijkl
+// Tests a reordering needing a single left move. Even though 'h is in the
```
95-martin-orion: `'h` -> `h`

tests/src/tensor_test.cpp
```diff
@@ -311,9 +311,11 @@ TEST(TensorTest, LeftTwelveIndicesReordering) {
...
+// rightmost grouping, left_reorder() will check and see if it can perform the
+// reordering by making a left move with up to the 8th index from the left.
+// abcdefgh | ijkl
+// -> hbcdefg | aijk
```
95-martin-orion: The set of index names changed between the first and second lines here...

tests/src/tensor_test.cpp
alankao64: changed to 'h'

tests/src/tensor_test.cpp
```diff
@@ -311,11 +311,11 @@ TEST(TensorTest, LeftTwelveIndicesReordering) {
...
 
-// Tests a reordering needing a single left move. Even though 'h is in the
-// rightmost grouping, left_reorder() will check and see if it can perform the
-// reordering by making a left move with up to the 8th index from the left.
-// abcdefgh | ijkl
-// -> hbcdefg | aijk
+// Tests a reordering needing a single left move. Even though index 'h' should
+// be in the rightmost grouping, left_reorder() will check and see if it can
+// perform the reordering by making a left move with up to the 8th index from
+// the left. abcdefgh | ijkl
```
95-martin-orion: Oh no, the auto-formatter :(

95-martin-orion: After this, it's ready to merge as far as I'm concerned.

alankao64: used `// clang-format off` and `// clang-format on`

PR #242, opened 2019-11-27
Added Rochester circuits, grid, and contraction

benjaminvillalonga:


benjaminvillalonga:
These were added with #252.

PR #248, opened 2019-12-11
Fix ordering_data_to_contraction_ordering

s-mandra:


tests/src/contraction_utils_test.cpp
```diff
@@ -442,24 +454,73 @@ TEST(OrderingParserTest, ParserFailures) {
...
+  try {
+    ordering_data_to_contraction_ordering(input, &ordering);
+    FAIL() << "Expected ordering_data_to_contration_ordering to throw an "
+              "expection.";
+  } catch (const std::string& msg) {
+    EXPECT_THAT(msg,
+                testing::HasSubstr(
+                    "Cut values must be comma-separated ints, e.g. (0,1,3)."));
+  }
   // Spaces are not allowed in the value list.
```
95-martin-orion: @s-mandra, does the ordering-file parser intelligently remove spaces now? If so, we can probably remove the `Spaces are not allowed in the value list` test case.

src/evaluate_circuit.cpp
```diff
@@ -136,10 +136,8 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
   std::list<ContractionOperation> ordering;
   try {
     ordering_data_to_contraction_ordering(*input, &ordering);
-  } catch (const std::string& err_msg) {
-    throw ERROR_MSG(
-        "Failed to call ordering_data_to_contraction_ordering(). Error:\n\t[",
-        err_msg, "]");
+  } catch (...) {
+    std::rethrow_exception(std::current_exception());
```
95-martin-orion: Do we need the try-catch block here anymore? If `ordering_data_to_contraction_ordering` generates an exception and we don't append anything to it, it would be simpler just to allow the original exception to be thrown directly.

s-mandra: Agreed (just removed rethrow)

tests/src/contraction_utils_test.cpp
s-mandra: `QflexOrdering::load` already takes care of reformatting everything to be nicely processed by other functions (including removing extra spaces). At the moment, `OrderingParserTest` is not testing `QflexOrdering::load`: perhaps, it would make more sense to actually use `QflexOrdering::load` and check if the output has been properly processed. Let me change this test.

s-mandra: Fixed.

PR #249, opened 2019-12-17
Python pypi package

alexandrupaler:
WIP


alexandrupaler:
Would you agree with reorganising the directories of the code? For preparing a python package it would make sense to move the python and src folders into a qflex folder and rename them something like
`qflex\cirq_interface` and `qflex\simulator`. Tests, scripts and everything else we could keep outside of the stuff we want to deliver. However, such a restructuring would mean to edit compilation scripts...

s-mandra:
> Would you agree with reorganising the directories of the code? For preparing a Python pypi package it would make sense to move the python and src folders into a qflex folder and rename them something like
> `qflex\cirq_interface` and `qflex\simulator`. Tests, scripts and everything else we could keep outside of the stuff we want to deliver. However, such a restructuring would mean to edit compilation scripts...
> 
> @s-mandra @95-martin-orion

Let me say that I would prefer to keep the structure as it is to avoid to rewrite all the scripts + docs. Is there any major issue to keep the structure as it is to build the pypi package?

alexandrupaler:
It would greatly simplify the preparation. Having all the code that needs to be delivered in a single folder, is easier to maintain. For example, the `src` and `python` folders contain both code but of different languages. For example, at Cirq they package only the `cirq` folder where everything necessary for the distribution exists. Other frameworks do the same.

95-martin-orion:
> For example, at Cirq they package only the `cirq` folder where everything necessary for the distribution exists.

Is this possible with qFlex, given that the Python code depends on the C++ libraries?

I'm also in favor of keeping the current directory structure, since it emphasizes that there are two ways to interface with the qFlex code (either through Python or C++). However, if we wanted to make the interface exclusively Python-based, nesting the C++ code under the Python directory might be reasonable.

Do we know how other pybind-based repos handle this? I was able to find [this Github issue],(https://github.com/pybind/pybind11/issues/1430) and subsequently [this repo](https://github.com/matplotlib/mplcairo) which uses pybind and installs via pip - if you have other examples, let me know.

alexandrupaler:
I am arguing in favour of moving the python and the c++ code below another folder called qflex. It would resemble a module structure for Python (making it similar to [ProjectQ](https://github.com/ProjectQ-Framework/ProjectQ/tree/develop/projectq) and the `backends` folder there.

I will try to finalise the package without code reorganisation, but due to some difficulties I ended up suggesting this potential solution because of [this](https://www.benjack.io/2018/02/02/python-cpp-revisited.html).

alexandrupaler:
Only missing issue is the config files are not the in the package. I do not have any clue how those should be delivered. Suggestions very welcome.

src/evaluate_circuit.cpp
```diff
@@ -188,6 +188,7 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
         }
         std::size_t tensor_size = 1;
         for (const auto& [index, dim] : index_dim) {
+          index + "do_nothing";
```
95-martin-orion: If index isn't needed in the loop, prefer:
```
for (const auto& pair : index_dim) {
  tensor_size *= pair.second;
}
```

alexandrupaler:
> ERROR: Failed building wheel for qflexcirq

Do you have wheel installed in the venv? I received this error before, and added wheel to requirements.txt. Am not sure why this happened now.

s-mandra:
> > ERROR: Failed building wheel for qflexcirq
> 
> Do you have wheel installed in the venv? I received this error before, and added wheel to requirements.txt. Am not sure why this happened now.

I'm having the same problem: could it be the format fix I did? (even though I would be surprised by that ..)

alexandrupaler:
> I'm having the same problem:

What happens when you write `pip3 install wheel`?

s-mandra:
> > I'm having the same problem:
> 
> What happens when you write `pip3 install wheel`?

Now it works. Just changed conda environment and it works well with:
```
pip3 install git+https://github.com/ngnrsaa/qflex@setuppypi
```

alexandrupaler:
I am also receiving sometimes incompatibility errors due to the packages in existing environments. I am testing with blank venvs.

tests/python/simulation_test.py
```diff
@@ -12,17 +12,20 @@
...
-import python.cirq_interface.qflex_order as qorder
+from qflexcirq import utils
+from qflexcirq import qflex
 
-import python.utils as qflexutils
+# import python.interface.qflex_simulator as qsim
+# import python.interface.qflex_virtual_device as qdevice
+# import python.interface.qflex_grid as qgrid
+# import python.interface.qflex_circuit as qcirc
+# import python.interface.qflex_order as qorder
```
s-mandra: If not required, perhaps we can remove these comments.

PR #250, opened 2019-12-20
Update install.md

alankao64:


docs/install.md
```diff
@@ -12,7 +12,7 @@ $ sudo apt-get install autoconf automake make g++ libgsl-dev libgslcblas0
 ```
 followed by the installation of required python modules:
 ```
-# python3 -m pip install -r scripts/requirements.txt
+$ python3 -m pip install -r scripts/requirements.txt
```
s-mandra: Perhaps, we want to add `--user` after `pip install` because we are switching from `root` to `non-root` user.

PR #251, opened 2020-01-08
Quick fix for LGTM.

s-mandra:
Remove unused check.


s-mandra:
> Makes sense - `x` and `y` are of type `size_t` if I'm reading this correctly.

Yes, that's correct. Thanks!

PR #252, opened 2020-01-17
Reduce memory requirements

benjaminvillalonga:
Actual tensor sizes are taken into account for allocating space for the contraction, as opposed to ranks and the max bond dimension of the network. This reduces the memory requirements for contracting tensor networks with heterogeneous bond dimensions.


benjaminvillalonga:
I have also added some new circuits that I should have committed separately. They are some of the circuits that would have allocated more memory than needed.

src/contraction_utils.cpp
```diff
@@ -54,16 +55,19 @@ ContractionData ContractionData::Initialize(
         std::vector<std::vector<std::string>>((*tensor_grid)[i].size());
     for (std::size_t j = 0; j < (*tensor_grid)[i].size(); ++j) {
       grid_indices[i][j] = (*tensor_grid)[i][j].get_indices();
-      for (const std::size_t dim : (*tensor_grid)[i][j].get_dimensions()) {
-        bond_dim = std::max(bond_dim, dim);
+      for (const std::string index : (*tensor_grid)[i][j].get_indices()) {
```
s-mandra: It's better to use constant references to `std::string`. I would modify this line as:
```
for(const std::string &index : ...)
```

s-mandra: It's better to use constant references to std::string. I would modify this line as:
```
for(const std::string &index : ...)
```

src/contraction_utils.cpp
```diff
@@ -74,23 +78,34 @@ ContractionData ContractionData::Initialize(
         patch_indices[op.expand.id] =
             _vector_subtraction(_vector_union(indices_a, indices_b),
                                 _vector_intersection(indices_a, indices_b));
-        if (data.patch_rank_[op.expand.id] <
-            patch_indices[op.expand.id].size()) {
-          data.patch_rank_[op.expand.id] = patch_indices[op.expand.id].size();
+        std::size_t new_size(1);
+        for (auto index : patch_indices[op.expand.id]) {
```
s-mandra: Same here. I would use:
```
for(const auto &index : ...)
```
since index is constant `std::string`.

src/contraction_utils.cpp
```diff
@@ -74,23 +78,34 @@ ContractionData ContractionData::Initialize(
...
           for (auto& tensor : op.cut.tensors) {
             auto& indices = grid_indices[tensor[0]][tensor[1]];
             const std::string copy_name = cut_copy_name(op.cut.tensors, side);
-            cut_copy_rank[copy_name] = indices.size();
             ++side;
+            // Update indices so it keeps track of cuts so that tensor sizes and
+            // ranks are not overestimated.
             indices = _vector_subtraction(indices, {cut_index});
+            std::size_t new_size(1);
+            for (auto index : indices) {
```
s-mandra: Same here:
```
for(const auto &index : indices)
```

src/contraction_utils.cpp
```diff
@@ -103,10 +118,15 @@ ContractionData ContractionData::Initialize(
         patch_indices[op.merge.target_id] =
             _vector_subtraction(_vector_union(indices_a, indices_b),
                                 _vector_intersection(indices_a, indices_b));
-        if (data.patch_rank_[op.merge.target_id] <
-            patch_indices[op.merge.target_id].size()) {
-          data.patch_rank_[op.merge.target_id] =
-              patch_indices[op.merge.target_id].size();
+        std::size_t new_size(1);
+        for (auto index : patch_indices[op.merge.target_id]) {
```
s-mandra: Same here:
```
for(const auto &index : ...)
```

src/contraction_utils.cpp
```diff
@@ -171,43 +193,38 @@ ContractionData ContractionData::Initialize(
...
-      throw ERROR_MSG("Failed to call Tensor(). Error:\n\t[", err_msg, "]");
+  {
+    std::size_t patch_pos = 1;
+    for (const auto& swap_size : unique_sizes) {
+      try {
+        data.scratch_.push_back(Tensor({""}, {swap_size}));
+      } catch (const std::string& err_msg) {
+        throw ERROR_MSG("Failed to call Tensor(). Error:\n\t[", err_msg, "]");
+      }
+      data.scratch_map_[result_space(swap_size)] = patch_pos;
```
s-mandra: Once @95-martin-orion suggested me not to use underscores at the beginning of variables. Similarly, should we change `scratch_map_` to just `scratch_map`?

src/contraction_utils.h
```diff
@@ -179,11 +179,11 @@ class ContractionData {
...
 
-  // Map of patch IDs/index names to scratch locations.
-  std::unordered_map<std::string, std::size_t> scratch_map_;
+  // Map of patch IDs/index names to scratch locations on scratch_.
+  std::unordered_map<std::string, size_t> scratch_map_;
 
-  // Max rank of each patch generated during contraction.
-  std::unordered_map<std::string, std::size_t> patch_rank_;
+  // Max size of each patch generated during contraction.
+  std::unordered_map<std::string, size_t> patch_max_size_;
```
s-mandra: Same here: `size_t` --> `std::size_t`.

src/tensor.h
```diff
@@ -164,8 +164,8 @@ class Tensor {
 
   /**
    * Get the allocated space of the Tensor. This tensor can be resized to
-   * any shape with total dimension less than this value.
-   * @return std::size_t with the capacity of the Tensor.
+   * any shape with total dimension (size) less than this value.
+   * @return int with the capacity of the Tensor.
```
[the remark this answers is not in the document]
s-mandra: Yes. I think the merge has just messed up this comment.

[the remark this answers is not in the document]
s-mandra: Yes. I think the merge just messed up with this comment.

s-mandra:
It looks good to me. After fixing all the nits, I'm ok with merging this PR.

src/contraction_utils.cpp
95-martin-orion: Trailing underscores are conventionally used to denote private members of a class - this naming looks OK to me.

src/contraction_utils.cpp
```diff
@@ -118,33 +138,35 @@ ContractionData ContractionData::Initialize(
...
   }
-  std::size_t max_size = static_cast<std::size_t>(pow(bond_dim, max_rank));
 
   // General-purpose scratch space (primarily used for tensor reordering).
   allocated_space += max_size;
 
   // "Swap tensor" space, used to store operation results.
-  for (std::size_t rank = 1; rank <= max_rank; ++rank) {
-    allocated_space += static_cast<std::size_t>(pow(bond_dim, rank));
+  {
```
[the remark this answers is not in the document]
benjaminvillalonga: Oops, that was inherited from a previous version.

src/contraction_utils.h
```diff
@@ -158,11 +159,10 @@ class ContractionData {
     return std::string(buffer, len);
   }
 
-  // Gets the index of result scratch space of the given rank.
-  static std::string result_space(std::size_t rank) {
+  // Gets the index of result scratch space of the given size.
+  static std::string result_space(std::size_t size) {
     char buffer[64];
```
s-mandra: I would avoid the use of buffers (it would be better to use `std::strings` and `std::stringstream`. I can work on this.

s-mandra: Instead of buffers, I'm directly using the `qflex::utils::concat` function, which is now in `src/utils.h`.

benjaminvillalonga:
Some examples on memory usage and runtimes on a single core of my laptop:
sycamore depth 6: 904 KB and 25 s
rochester depth 20: 2.4 GB and 70 s
bristlecone-48 depth 24: 52.8 MB and 7.5 s (with the current fidelity settings on the ordering file)
bristlecone-70 depth 20: 7.3 MB and 191 s (with the current fidelity settings)

95-martin-orion:
> Some examples on memory usage and runtimes on a single core of my laptop: {...}
> bristlecone-70 depth 20: 7.3 MB and 191 s (with the current fidelity settings)

Was this run 7.3MB or 7.3GB? Either way, all of these numbers look good to me.

benjaminvillalonga:
> Was this run 7.3MB or 7.3GB? Either way, all of these numbers look good to me.

It was 7.26 MB. Great, merging!

PR #253, opened 2020-01-17
Track memory usage in real-time (depends on: #257).

s-mandra:


s-mandra:
For now, it works for linux only. It uses `signal`s and `/proc/self/status`, which are not available in neither Windows nor MacOSX. Any ideas?

s-mandra:
> For other operating systems:
> 
> * Windows has [`GetProcessMemoryInfo`](https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo).
> * Mac doesn't appear to have good support for this, although it seems like the [UNIX `time` command](https://stackoverflow.com/questions/12476804/how-i-can-get-the-peak-memory-on-mac-os) may return the required info if you can get the PID.
> 
> I'm okay with leaving these as TODO items for now as long as this doesn't cause the other OSes to crash when logging memory usage.

Fixed the code so that tracking is only enabled for `linux` systems.

s-mandra:
For some reasons, `ualarm` doesn't work properly for `track_memory_milliseconds >= 1000`. The problem doesn't appear for `MacOSX`.

s-mandra:
> For some reasons, `ualarm` doesn't work properly for `track_memory_milliseconds >= 1000`. The problem doesn't appear for `MacOSX`.

It seems that actually `ualarm` only accept intervals smaller than 1 second.

src/memory.h
```diff
@@ -5,40 +5,73 @@
...
-    line = line.substr(line.find_first_of("0123456789"), std::size(line));
+    // Read file and get values
+    std::string line, m_used, m_peak;
+    while(std::getline(in, line)) {
 
-    // Strip everything after the first space and get memory usage
-    const std::string m_used = utils::readable_memory_string(
-        std::stod(line.substr(0, line.find(' '))) * 1024);
+      if(line.find("VmHWM", 0) != std::string::npos) m_peak = get_value(line);
+      else if(line.find("VmRSS", 0) != std::string::npos) m_used = get_value(line);
```
95-martin-orion: I don't think `else` is needed here (assuming we want to return both current and peak memory usage).

95-martin-orion: Disregard; was reading an outdated version.

s-mandra:
> For other operating systems:
> 
> * Windows has [`GetProcessMemoryInfo`](https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo).
> * Mac doesn't appear to have good support for this, although it seems like the [UNIX `time` command](https://stackoverflow.com/questions/12476804/how-i-can-get-the-peak-memory-on-mac-os) may return the required info if you can get the PID.
> 
> I'm okay with leaving these as TODO items for now as long as this doesn't cause the other OSes to crash when logging memory usage.

It now works for both `linux` and `macOSX`. The portion of the code related to `win64` is commented because not tested.

PR #254, opened 2020-01-18
Fix memory cmd argument.

s-mandra:
At the moment, `--memory` requires an integer representing the maximum amount of memory (in bytes) allowed during the simulation. This PR fixes the issues by allowing human-readable strings as `10GB`.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -16,6 +16,6 @@ std::string readable_memory_string(double memory) {
   return concat(memory, suffix[scale]);
 }
 
-}
+}  // namespace qflex::utils
 
 #endif
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -5,6 +5,38 @@
 
 namespace qflex::utils {
 
+// Convert more readable strings into 'memory' bytes.
+std::size_t from_readable_memory_string(std::string memory) {
+
+  // Remove any special character
+  memory = std::regex_replace(memory, std::regex("[^)(\\s\\ta-zA-Z0-9_.,-]"), "");
+
+  // Convert tabs to spaces
+  memory = std::regex_replace(memory, std::regex("[\\t]"), " ");
+
+  // Remove all spaces
+  memory = std::regex_replace(memory, std::regex(" "), "");
+
+  // Convert all letters to uppercase
+  for(auto &c: memory) c = std::toupper(c);
+
+  std::size_t m = std::stoll(memory.substr(0, memory.find_first_not_of("0123456789")));
+
+  {
+    // Find the first non-number and get the suffix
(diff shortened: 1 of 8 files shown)


s-mandra:
For now, it works for linux only. I'm looking for something similar for MacOSX, but I still haven't found a solution.

s-mandra:
> This change looks fine - it shouldn't be causing the test failures. Maybe a sync is needed?

Yeah .. not sure why. Looking into it.

s-mandra:
> This change looks fine - it shouldn't be causing the test failures. Maybe a sync is needed?

The problem seems to be related to some issues with `Docker` (it compiles and run all the tests in my linux machine).

s-mandra:
> > This change looks fine - it shouldn't be causing the test failures. Maybe a sync is needed?
> 
> The problem seems to be related to some issues with `Docker` (it compiles and run all the tests in my linux machine).

Found the error: when called directly, `qFlex` library tries to access `qflex::global::memory_limit` which is uninitialized (because, for regular calls, `docopt` does it). In `Docker` containers, `qflex::global::memory_limit` is set to zero, which makes the simulation throw an error. Fixing it.

src/utils.cpp
```diff
@@ -18,20 +18,29 @@ std::size_t from_readable_memory_string(std::string memory) {
...
-    else if(suffix == "KB") m *= std::pow(1024, 1);
-    else if(suffix == "B")  m *= std::pow(1024, 0);
-    else if(suffix == "")   m *= std::pow(1024, 0);
-    else throw ERROR_MSG("Suffix ", suffix, " not recognized.");
+    const std::string suffix = memory.substr(
+        std::min(memory.find_first_not_of("0123456789"), std::size(memory)),
+        std::size(memory));
+
+    if (suffix == "GB")
+      m *= std::pow(1024, 3);
```
95-martin-orion: Nit: depending on compiler, `m << 30` may be slightly faster - not that it makes much difference, given that this is only called once at startup. If you think the current format will be clearer for readers, please leave as-is.

s-mandra: Agreed. However, modern compilers should understand that `std::pow` Is a `constexpr` and eventually compute the result in compile time and reuse it in runtime. Probably, the best would be to provide `const` vars to be 100% sure that these values are not recompiled each time.

s-mandra: What’s your opinion about using [user defined literals](https://stackoverflow.com/questions/39622508/how-to-write-a-literal-constant-like-size-t-s-16-mbyte/39625300)?

95-martin-orion: This is probably more effort (both for developers and users trying to understand the code) than it's worth. All things considered, there should be no issues using the `std::pow` format.

PR #256, opened 2020-01-22
Fix docopt default values.

s-mandra:
To avoid confusions, default values for `docopt` are now read from `qflex::global`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4,12 +4,13 @@
 #include "global.h"
 #include "grid.h"
 #include "input.h"
+#include "utils.h"
 #include "memory.h"
 #include "ordering.h"
 
 static const char VERSION[] = "qFlex v0.1";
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -23,13 +25,15 @@ tensor network, CPU-based simulator of large quantum circuits.
...
+        qflex::global::verbose, R"(].
+    -m,--memory=<memory_limit>             Memory limit [default: )",
+        qflex::utils::readable_memory_string(qflex::global::memory_limit), R"(].
     --initial-conf=<initial_conf>          Initial configuration.
     --final-conf=<final_conf>              Final configuration.
     --version                              Show version.
-
-)";
+)")
+        .c_str();
```
95-martin-orion: This `.c_str()` probably isn't needed.

PR #260, opened 2020-01-24
Cleanup code & Improvements.

s-mandra:
This PR tries to remove all pointers (when possible) with (const, if possible) references and cleanup the code.

### DIFF
--- a/src/evaluate_circuit.cpp
+++ b/src/evaluate_circuit.cpp
@@ -19,13 +19,10 @@
 namespace qflex {
 
 // Gets the position in the output state vector of the qubit at tensor_pos.
-std::size_t find_output_pos(const QflexInput* input,
-                            std::vector<std::size_t> tensor_pos) {
-  if (input == nullptr) {
-    throw ERROR_MSG("Input must be non-null.");
-  }
-  std::size_t pos = (tensor_pos[0] * input->grid.J) + tensor_pos[1];
-  for (const auto off_pos : input->grid.qubits_off) {
+std::size_t find_output_pos(const QflexInput& input,
+                            const std::vector<std::size_t>& tensor_pos) {
+  std::size_t pos = (tensor_pos[0] * input.grid.J) + tensor_pos[1];
+  for (const auto off_pos : input.grid.qubits_off) {
     if (off_pos[0] < tensor_pos[0]) {
       --pos;
     } else if (off_pos[0] == tensor_pos[0] && off_pos[1] < tensor_pos[1]) {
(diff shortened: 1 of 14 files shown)


src/evaluate_circuit.cpp
```diff
@@ -36,65 +33,50 @@ std::size_t find_output_pos(const QflexInput* input,
...
-    throw ERROR_MSG("Input must be non-null.");
-  }
-  if (final_qubits == nullptr) {
-    throw ERROR_MSG("Final qubits must be non-null.");
-  }
-  if (output_states == nullptr) {
-    throw ERROR_MSG("Output states must be non-null");
-  }
+    const QflexInput& input, const std::list<ContractionOperation>& ordering,
+    std::vector<std::vector<std::size_t>>& final_qubits,
```
95-martin-orion: Pointers are preferred for non-const pass-by-reference arguments, as this clarifies at the call site which arguments can be mutated by the function. For example:

```
void foo(int& val) { val++; }
void bar(int* val) { (*val)++; }
...
int x = 1;
foo(x);  // without reading foo, it's unclear that foo can modify 'a'
bar(&x); // explicit pass-by-reference: bar has clear access to modify 'a'
```
This doesn't apply to const arguments - `const QflexInput& input` is fine.

s-mandra: Agreed. Let me say that the best would be (and let me know if you agree):

1. If function `foo` requires a `const` single variable `x`, pass `x` as reference,
2. If function `foo` requires a single (but already initialized, so it will be modified in `foo`) variable `x`, pass `x` as pointer,
3. If function `foo` requires a single uninitialized (which will be initialized in `foo`), remove `x` from the arguments so that `foo` can return it,
4. If function `foo` requires multiple uninitialized (which will be initialized in `foo`) `x`, `y` ..., remove `x`, `y`, ... and return them as `std::tuple` (very rare in our case),

Mixed case will be treated accordingly. What do you think @95-martin-orion?

95-martin-orion: I'm mostly on board with this plan. Responses by case:

1. Looks good.
2. This can apply for multiple pre-initialized arguments as well: `foo(A* x, B* y, ...)`
3. Works as long as the function previously returned `void`; otherwise see case 4.
4. Depending on the situation, it may be better to define a struct for the return type (e.g., if there are many such args for a single function). Since these are rare, we can go through them on a case-by-case basis.

s-mandra:
Too many modifications has been done to other PR (see #267) to be compatible with this PR. I'm closing it.

PR #265, opened 2020-01-27
Improve flattening of 3D tensor.

s-mandra:


src/read_circuit.cpp
```diff
@@ -545,7 +545,7 @@ void circuit_data_to_tensor_network(
 
 // TODO: add tests for this function. Optimize contraction procedure; it uses
 // far more memory than needed currently. Compactify code. Use smart pointers
-// where possible.
+// where possible. Improved.
```
95-martin-orion: Also delete the `Optimize contraction...` sentence - this PR does the requested optimization.

PR #268, opened 2020-02-03
Improve timing and warning messages.

s-mandra:


src/evaluate_circuit.cpp
```diff
@@ -109,33 +109,35 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
...
+  // Print difference in time
+  auto print_diff_time = [&t0, &t1](const std::string& msg) {
+    const double delta_time_ms =
+        std::chrono::duration_cast<std::chrono::milliseconds>(
+            (t1 = std::chrono::steady_clock::now()) - t0)
+            .count() /
+        1000.;
+    std::cerr << WARN_MSG(msg, delta_time_ms, "s") << std::endl;
+    t0 = std::move(t1);
+  };
```
95-martin-orion: I think it would be cleanest to handle this with a separate class, especially since a similar structure is useful in other places:
```
// pseudocode
class Stopwatch {
  time_point start, latest;

  start() { start = latest = now() }

  update_latest_and_log_delta(message) {
    delta = now() - latest
    log message with delta
    // logging takes time, so we call "now()" again
    latest = now()
  }

  log_total() {
    delta = now() - start
    log message with delta
  }
}
```

src/evaluate_circuit.cpp
```diff
@@ -109,33 +109,35 @@ std::vector<std::pair<std::string, std::complex<double>>> EvaluateCircuit(
...
 
-  if (global::verbose > 0) {
-    t1 = std::chrono::high_resolution_clock::now();
-    time_span =
-        std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0);
-    std::cerr << "Time spent making contraction ordering: " << time_span.count()
-              << "s" << std::endl;
-  }
+  if (global::verbose > 0)
+    print_diff_time("Time spent making contraction ordering: ");
```
95-martin-orion: As it is, it's unclear that this modifies `t0` - especially since `t0` isn't one of the function arguments.

src/utils.cpp
```diff
@@ -5,6 +5,16 @@
...
 namespace qflex::utils {
 
+// Return date and time in the representation: Thu Aug 23 14:55:02 2001
+std::string get_date_time() {
+  std::time_t t = std::time(nullptr);
+  if (char mbstr[100];
+      std::strftime(mbstr, sizeof(mbstr), "%c", std::localtime(&t)))
+    return mbstr;
+  else
+    return "n/a";
```
95-martin-orion: This string should indicate that this is an time-readout error.

src/evaluate_circuit.cpp
s-mandra: I'll add an extra function in utils.

s-mandra: Agreed.

src/evaluate_circuit.cpp
s-mandra: Agreed, this should be solved once an object `Stopwatch` is created.

src/stopwatch.h
```diff
@@ -0,0 +1,99 @@
...
+   */
+  void stop();
+
+  /**
+   * Reset Stopwatch
+   */
+  void reset();
+
+  /**
+   * Compute the difference in time between clock::now() and the last _split.
```
95-martin-orion: This comment should note that the function also updates `_split`.

PR #269, opened 2020-02-03
Improve tensor class (depends on: #268)

s-mandra:
Copy/Move constructors and copy/move assignment operators are now 
consistent with their function. Now, the following code:
```
std::vector<std::vector<Tensor>> tensors(...);
[...]
auto copy_tensors = tensors;
```
works without segmentation fault.

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -1035,7 +1035,8 @@ void multiply(Tensor& A, Tensor& B, Tensor& C, s_type* scratch_copy) {
     t1 = std::chrono::high_resolution_clock::now();
     time_span =
         std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0);
-    std::cerr << "Time preparing variables: " << time_span.count() << "s\n";
+    std::cerr << WARN_MSG("Time preparing variables: ", time_span.count(), "s")
+              << std::endl;
   }
 
   // Check.
@@ -1060,8 +1061,10 @@ void multiply(Tensor& A, Tensor& B, Tensor& C, s_type* scratch_copy) {
     t1 = std::chrono::high_resolution_clock::now();
     time_span =
         std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0);
-    std::cerr << "R " << time_span.count() << " s\n";
-    std::cerr << "Time reordering A: " << time_span.count() << "s\n";
+    std::cerr << WARN_MSG("R ", time_span.count(), "s") << std::endl;
+    ;
+    std::cerr << WARN_MSG("Time reordering A: ", time_span.count(), "s")
+              << std::endl;
     t0 = std::chrono::high_resolution_clock::now();
   }
 
@@ -1077,8 +1080,9 @@ void multiply(Tensor& A, Tensor& B, Tensor& C, s_type* scratch_copy) {
     t1 = std::chrono::high_resolution_clock::now();
     time_span =
         std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0);
-    std::cerr << "R " << time_span.count() << " s\n";
-    std::cerr << "Time reordering B: " << time_span.count() << "s\n";
+    std::cerr << WARN_MSG("R ", time_span.count(), "s") << std::endl;
+    std::cerr << WARN_MSG("Time reordering B: ", time_span.count(), "s")
+              << std::endl;
     t0 = std::chrono::high_resolution_clock::now();
   }
 
@@ -1119,8 +1123,9 @@ void multiply(Tensor& A, Tensor& B, Tensor& C, s_type* scratch_copy) {
     t1 = std::chrono::high_resolution_clock::now();
     time_span =
         std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0);
-    std::cerr << "M " << time_span.count() << " s\n";
-    std::cerr << "Time multiplying A*B: " << time_span.count() << "s\n";
+    std::cerr << WARN_MSG("M ", time_span.count(), "s") << std::endl;
+    std::cerr << WARN_MSG("Time multiplying A*B: ", time_span.count(), "s")
+              << std::endl;
     t0 = std::chrono::high_resolution_clock::now();
   }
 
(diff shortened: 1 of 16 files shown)


src/tensor.cpp
```diff
@@ -178,16 +150,51 @@ Tensor::Tensor(std::vector<std::string> indices,
...
+Tensor::Tensor(const Tensor& other)
+    : _indices{other._indices},
+      _dimensions{other._dimensions},
+      _index_to_dimension{other._index_to_dimension},
+      _capacity{other._capacity} {
+  // Allocate space
+  _data = new s_type[_capacity];
+
+  // Copy data
+  std::copy(other._data, other._data + other._capacity, _data);
```
95-martin-orion: `other._data` is often smaller than its `capacity`, which would make this do more work than necessary. Use `other.size()` instead of `other._capacity`.

src/tensor.h
```diff
@@ -23,6 +23,10 @@
 #include <cblas.h>
 #endif
 
+#ifdef _OPENMP
+#include <omp.h>
```
95-martin-orion: Is this needed outside of `tensor.cpp`? Otherwise it probably should stay in that file.

src/tensor.cpp
```diff
@@ -178,16 +150,51 @@ Tensor::Tensor(std::vector<std::string> indices,
   _data = data;
 }
 
-Tensor::Tensor(const Tensor& other) { _copy(other); }
+Tensor::Tensor(const Tensor& other)
```
95-martin-orion: Does `*this = other` work here? This method has slightly different behavior from the equals operator, which could lead to _very_ confusing errors.

src/tensor.cpp
```diff
@@ -95,31 +91,6 @@ void Tensor::_clear() {
   }
 }
 
-void Tensor::_copy(const Tensor& other) {
```
95-martin-orion: What is the motivation behind removing this method? If performance is a concern, converting to inline (e.g. `inline void Tensor::_copy(...)`) should have the same effect.

src/tensor.h
s-mandra: It depends. I prefer to have all the `include`s in the header file, but I'm ok to keep it in `tensor.cpp`.

src/tensor.cpp
s-mandra: > Does `*this = other` work here? This method has slightly different behavior from the equals operator, which could lead to _very_ confusing errors.

No. `*this = other` would call `operator=()`, which uses the code originally in `_copy`. `Tensor(const Tensor&)` is called from a direct instantiation like `auto tensor = other`. In this case, it makes more sense to have `tensor` identical to `other` (in terms of stored data), so every member in `other` is just cloned in `tensor`.

src/tensor.cpp
s-mandra: `_copy` was used in both `operator=(const Tensor&)` and in `Tensor(const Tensor&)`. However, `_copy` works in a slightly different way one would expect for a copy constructor (a copy constructor should clone the original tensor), so I needed to fix the copy constructor for tensor. Now, the original code for `_copy` has been moved to `operator(const Tensor&)` while the proper cloning of the object has been added in `Tensor(const Tensor&)`. Without my modification, code like the following:
```
std::vector<std::vector<Tensor>> tensors(...);
/*
  operations on tensors
*/
auto copy_of_tensors = tensors;
```
would give segmentation fault. Since the old `_copy` is only used in `operator=(const Tensor&)`, I've just decided to move the code there to avoid confusion.

src/tensor.h
95-martin-orion: In general, it's best to put as much as possible in the .cpp file - any `include` in the .h file must be included by every file that references that header, which will slow down compilation.

src/tensor.cpp
95-martin-orion: I think the root issue here is this section in the original `_copy()` method:
```
_capacity = other.size();	
_data = new s_type[_capacity];
```
This is probably meant to reduce memory usage by limiting the copied tensor to the size of the original's data (rather than its capacity). However, a bristlecone_48 run with the `track_memory` flag shows minimal difference between this behavior and keeping the original capacity. By that measure it should be OK to clone the capacity of `other`, which should give the expected behavior in the constructor.

s-mandra: > By that measure it should be OK to clone the capacity of `other`, which should give the expected behavior in the constructor.

I agree. Cloning should give an exact copy (including the original capacity).

95-martin-orion: I forgot to give my recommendation here: we should keep `_copy()` and just replace the lines above with
```
_capacity = other.capacity();	
_data = new s_type[_capacity];
```
since it doesn't appear to impact memory usage. (I'm mostly pushing for this to avoid having different behavior in the copy-constructor and equals operator.)

s-mandra: Moved back `Tensor::_copy()`. Made some improvements to have:
```
std::vector<std::vector<Tensor>> tensors(size_x, std::vector<Tensor>(size_y);
```
working.

src/tensor.h
```diff
@@ -70,21 +60,10 @@ class Tensor {
...
-  /**
-   * Creates a Tensor. A pointer to the data is passed.
-   * @param indices std::vector<std::string> with the names of the indices in
-   * order.
-   * @param dimensions std::vector<std::size_t> with the ordered dimensions of
-   * the indices.
-   * @param data pointer to the data of the tensor. It is responsibility of
-   * the user to provide enough allocated memory to store the Tensor.
-   */
-  Tensor(std::vector<std::string> indices, std::vector<std::size_t> dimensions,
```
s-mandra: This has been removed because i) it's never used and ii) it's unsafe (the destructor call `_clear()` which frees the memory for `_data`, but it shouldn't because the caller, and not the callee, should be responsible to clean-up the memory).

95-martin-orion: Good call. I think this was originally added for use in unit tests, but clearly we didn't actually need it.

src/tensor.cpp
95-martin-orion: Looks like `copy` is still using `_capacity = other.size();`

s-mandra: Thanks for catching that!

PR #271, opened 2020-02-06
Fixed build problems related to type conversion in cblas API.

DmitryLyakh:


s-mandra:
Hi @DmitryLyakh, would you mind to apply the following patch:
```
diff --git a/src/tensor.cpp b/src/tensor.cpp
index db1a623..0a24baa 100644
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -1024,8 +1024,8 @@ void _multiply_vv(const s_type* A_data, const s_type* B_data, s_type* C_data,
   }

   cblas_cdotu_sub(k, reinterpret_cast<const float*>(A_data), 1,
-                     reinterpret_cast<const float*>(B_data), 1,
-                     reinterpret_cast<openblas_complex_float*>(C_data));
+                  reinterpret_cast<const float*>(B_data), 1,
+                  reinterpret_cast<openblas_complex_float*>(C_data));
 }
 #endif
```
to pass the format check? Thanks!

PR #279, opened 2020-03-01
Memory limit can be specified in the Python interface.

s-mandra:


.gitignore
```diff
@@ -18,7 +18,7 @@
 todo.txt
 
 # Python cache
```
95-martin-orion: Is this still just for the Python cache?

src/pybind_main.cpp
```diff
@@ -80,6 +80,24 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
     // Get qFlex input
     qflex::QflexInput input;
 
+    // Set memory limit
+    if (options.contains("memory_limit")) {
+      const auto &memory_limit = options["memory_limit"];
+      try {
+        // Check if it can be converted to a long int
+        qflex::global::memory_limit = memory_limit.cast<long int>();
```
95-martin-orion: Would `size_t` instead of `long int` make sense here?

.gitignore
s-mandra: No, you're right (it's an old comment). I took the practice to call `__something__` anything that should be local, so I just changed it. Let me remove the comment.

src/pybind_main.cpp
s-mandra: The main reason why I'm using `long int` instead of `size_t` is because the user could pass a negative number that eventually would be casted to a huge positive number. I'm working on a workaround.

s-mandra:
@95-martin-orion, just did a quick fix to handle the case `options['memory_limit']` is neither a `py::int_` nor a `py::str`.

src/pybind_main.cpp
```diff
@@ -91,7 +91,8 @@ std::vector<std::pair<std::string, std::complex<double>>> simulate(
       } else if (py::isinstance<py::str>(memory_limit)) {
         qflex::global::memory_limit = qflex::utils::from_readable_memory_string(
             memory_limit.cast<std::string>());
-      }
+      } else
+        ERROR_MSG("'memory_limit' must be a valid.");
```
95-martin-orion: `must be a valid.` -> `must be a valid integer.`?

Issue #282, opened 2020-03-02
update qflexcirq documentation


alexandrupaler:
To include both methods of installing: 1) pypi 2) compile in git repo.

Is related to #281

Issue #281, opened 2020-03-02
scripts with consistent naming


alexandrupaler:
some of the scripts use "-" and others "_" in their file names. There is also "construct"/"create". Would it be fine to rename them with a consistent naming?

95-martin-orion:
This seems like a worthwhile cleanup step. To match the other files in the repo, I think "_" is preferable. We'll need to update references to these files as well (likely places are docs, config process, and other scripts).

PR #283, opened 2020-03-06
Minor fixes

alexandrupaler:
WIP. Addresses #281 #282 and includes documentation for setup.py


alexandrupaler:
Still WIP, but ready for feedback and review

setup.py
```diff
@@ -20,6 +20,36 @@
 unchanged. For that reason the AutoconfiCommand class is just a way to step into
 the extensions building process of setup.py and follow afterwards step-by-step
 the usual installation instructions.
+
+Mini HOWTO:
+
+* To upload to the pypi server the most simple way is to call
+```
+    python3 setup.by sdist
```
95-martin-orion: `setup.by` -> `setup.py`

setup.py
```diff
@@ -20,6 +20,36 @@
...
+performed by twine (in case needed `pip install twine`)
+
+* When the pypi server complains that the package already exists, most probably
+the version number below nees to be incremented.
+
+* Pypi allows uploading precompiled binaries packaged as wheels, but it is 
+a real pain to compile multi-platform. Because qflex is based on autotools, and
+the qflex-native compilation process is really streamlined, it makes no real 
+sense to complicate the installation. Users of qflexcirq will compile from
+sources on their machine QFlex.
```
95-martin-orion: I would reword this. There is definite value in being able to install from a wheel (at a minimum, it's a lot faster than compiling locally), we've just decided not to since it requires significant refactoring.

Maybe note that this is a possible area for future work? I wouldn't go so far as to call it a TODO though, since we don't really have plans TODO this :stuck_out_tongue:

setup.py
```diff
@@ -8,18 +8,57 @@
...
-build_ext, and we need to overload it in order to call the build chain
+build_ext, and we need to overload it in order to call the build chain.
 
-There are some so-called Extensions to handle Makefiles etc., but setuptools
-is such a voodoo piece of code difficult to debug, understand and use, that for
-our case it is easier and safer to call the automated chain from processes.
+The web has many examples on how to script so-called Extensions to handle 
+Makefiles etc.. However, setuptools is a voodoo piece of code difficult to 
+debug, understand and use. For Qflex it is easier and safer to call the 
+automated compilation chain using external subprocesses.
```
95-martin-orion: As above, consider rewording this. We don't want to discourage an ambitious contributor (or someone with lots of setuptools experience) from converting to a less client-heavy install process.

setup.py
```diff
@@ -8,18 +8,57 @@
...
 
 In order to not confuse the installation procedures between usual repository 
 clone and "pip install", the goal is to keep the toolchain as much as possible
-unchanged. For that reason the AutoconfiCommand class is just a way to step into
+unchanged. For that reason, the `AutoconfiCommand` class is used to step into
```
95-martin-orion: `AutoconfiCommand` -> `AutoconfigCommand` (missing `g`)

setup.py
alexandrupaler: I will reword. However, there is marginal value in having python call precompiled binaries for different platforms. Python may want, but will never be like java. The dependencies of Qflex are not trivial (e.g. blas) and it will be a huge pain to make sure that once binary compiled the users have the libs on their path. And the libs may have also different licences which may restrict packaging.

Issue #284, opened 2020-03-08
Confusing error message when copying rank-0 tensors


rmlarose:
I'm trying to simulate a circuit through the Cirq interface without reading in files. I'm able to run [qflex_cirq_example.py](https://github.com/ngnrsaa/qflex/blob/master/qflexcirq/qflex_cirq_example.py) with no issues. I modified this to construct the same circuit and grid used in the rectangular 2x2 example (see inline comments in code below).

I find that the circuit and grid don't have to be read in from files, but when the ordering is **not** read in from a file, I get a memory leak. 

The specific example I'm running (from the same directory as `qflex_cirq_example.py`) is below:

```python
import sys
import os
sys.path.insert(
    1, os.path.realpath(os.path.dirname(os.path.realpath(__file__)) + '/../'))

import cirq

from qflexcirq.interface import (qflex_circuit, qflex_simulator, qflex_order, qflex_virtual_device)
from qflexcirq.ordering import order_circuit_simulation


if __name__ == "__main__":
    # Same grid as in config/grid/rectangular_2x2.txt
    grid = "1 1\n1 1"
    device = qflex_virtual_device.QFlexVirtualDevice(qflex_grid=grid)
    qubits = device.get_indexed_grid_qubits()
    qreg = list(qubits.values())

    # Same circuit as in config/circuits/rectangular_2x2_1-2-1.txt
    circ = cirq.Circuit([cirq.ops.H.on_each(*qreg)])
    for _ in range(2):
        circ.append([cirq.ops.T.on_each(*qreg)])
        circ.append([cirq.ops.CNOT.on(qreg[i], qreg[i + 1]) for i in (0, 2)])
        circ.append([cirq.ops.T.on_each(*qreg)])
        circ.append([cirq.ops.CNOT.on(qreg[i], qreg[i + 2]) for i in (0, 1)])
    circ.append([cirq.ops.H.on_each(*qreg)])

    # Reading in the file works
    # order = qflex_order.QFlexOrder.from_existing_file("../config/ordering/rectangular_2x2.txt")

    # Computing the order (not reading in the file) throws an error
    order_list = order_circuit_simulation.circuit_to_ordering(circ)
    order = qflex_order.QFlexOrder(order_list)

    qflex_circ = qflex_circuit.QFlexCircuit(cirq_circuit=circ, device=device, qflex_order=order)
    sim = qflex_simulator.QFlexSimulator()

[...]

PR #285, opened 2020-03-09
Fix support for zero-cut orderings.

95-martin-orion:
Fixes an issue spotted in #284.

Previously, having zero cuts would cause this tool to spit out a blank contraction ordering. This PR adds an initial pass before applying cuts to ensure that zero-cut requests generate an ordering.

One side effect of this is that it is now possible for the auto-ordering tool to stop at zero cuts if adding the first cut fails to improve total cost (although such an occurrence is rare).


95-martin-orion:
I'll include tests in this PR as well - please hold off on review/merge until then.

95-martin-orion:
> I'll include tests in this PR as well - please hold off on review/merge until then.

Tests added, ready for review.

PR #287, opened 2020-03-11
Added sycamore_53 circuits with depths larger than 10

benjaminvillalonga:
Just added circuits for sycamore with depths beyond 10 cycles.


benjaminvillalonga:
> Looks good to me. Were these generated with the `qflexcirq/circuits/generator.py` utility?

No, these are the ones we had from supremacy before having the generator.

PR #295, opened 2020-07-24
Fix optimization flag on Power arch

mofeing:
Fixes a building error on Power architecture:
  - `-march=native` on x86
  - `-mcpu=native` on Power

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ AC_PROG_CXX
 CXXFLAGS="-O3  -std=c++17  -Wall -Wpedantic -Wno-unused-command-line-argument "$CXXFLAGS
 if test $(uname -m) == "x86_64"; then
   CXXFLAGS="-march=native "$CXXFLAGS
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -10,7 +10,12 @@ AC_LANG(C++)
 AC_PROG_CXX
 
 # Default CXXFLAGS.
-CXXFLAGS="-O3  -std=c++17  -march=native -Wall -Wpedantic -Wno-unused-command-line-argument "$CXXFLAGS
+CXXFLAGS="-O3  -std=c++17  -Wall -Wpedantic -Wno-unused-command-line-argument "$CXXFLAGS
+if test $(uname -m) == "x86_64"; then
```
95-martin-orion, line 14: For consistency, `-march=native` should be applied on _all_ non-Power architectures. `x86_64` is not the only arch that qFlex currently supports.

mofeing: Fixed!

PR #297, opened 2020-07-31
fix for ipcp build and new build option mkl

joveh:
fix building Intel MKL
- _multiply_MM() are also required when building with Inte MKL
- use generic type cast instead of openblas_complex_float also in
  the last occurance

fix CXXFLAGS for icpc; remove specific options for g++
icpc does not know -Wpedantic and -Wno-unused-command-line-argument

add configure option "--enable-mkl" to use Intel MKL
Intel MKL can also be used with g++ while icpc can also build with
OpenBASE. Thus, a new configure option is required

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -9,14 +9,20 @@ PREFIX = @prefix@
 PYBIND11 = @pybind11@
 PYTHON_TESTS = @python_tests@
 OPENMP = @openmp@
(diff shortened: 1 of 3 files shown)


95-martin-orion:
Added @s-mandra to review Makefile / configure changes. Overall this seems OK to me, but I'm not deeply familiar with the files being changed.

PR #299, opened 2023-12-19
Remove leading `$`s from example commands in README.md.

dlyongemallo:
The usual convention is to leave them off so users can just copy and paste them directly.


dlyongemallo:
It looks like my being no longer at Google broke the CLA bot. It no longer recognises that I've signed a CLA, but I can't sign a CLA because it claims I've already signed.

